PR #21, opened 2015-06-15
Add JSON wrapper class

djberg96:
This adds a BaseWrapper class (and one initial subclass) that encapsulates the raw JSON data returned from the rest calls. This allows us to use methods (including chained methods) instead of using raw string accessors, including snake_case aliases for camelized properties.

    # Example:
    vmm = VirtualMachineManager.new
    vm = vmm.get('your_vm')

    vm.properties.hardwareProfile.deploymentName
    vm.properties.hardware_profile.deployment_name # Same as above

I've also included some tests.


lib/azure/armrest/wrapper/base_wrapper.rb
```diff
@@ -0,0 +1,77 @@
...
+      # versions of the camelCase delegate methods.
+      def __setobj__(obj)
+        obj.methods(false).each{ |m|
+          if m.to_s[-1] != '=' # Must deal with nested ostruct's
+            res = obj.send(m)
+            __setobj__(res) if res.is_a?(OpenStruct)
+          end
+          snake = m.to_s.gsub(/(.)([A-Z])/,'\1_\2').downcase.to_sym
+          obj.instance_eval("alias #{snake} #{m}")
+        }
```
Fryguy: I fear this might get expensive.  This is explicitly one of the reasons we stopped using OpenStruct in the rest of our code base because they are expensive.

Fryguy:
Aside from the use of OpenStruct...looks great!

lib/azure/armrest/wrapper/base_wrapper.rb
djberg96: I thought about that, but I the json sent back per request is pretty small overall, so it shouldn't be an issue. If you have an alternative approach to this, I'd love to hear it.

Fryguy: I guess it really comes down to the usage of the methods that are being generated.  If we only call them once it turns out the overhead to dynamically define a method (which is what OpenStruct does) is not worth it.  In the ManageIQ code base, we just use method_missing and the overhead is removed.  If we are going to call the method like 100s of times however, then it's worth it to create the method.

See https://github.com/ManageIQ/manageiq/blob/master/lib/util/miq-hash_struct.rb for a class I had written that is like OpenStruct but uses method_missing instead of creating methods.  I remember when we implemented it, we got enormous performance increase, but that was also back on Ruby 1.8.7 :smile:   I had always intended on getting that back into Ruby proper somehow (like as an option to OpenStruct like `:generate_methods => false`) but never got around to it.

There may be an existing gem that does this kind of Struct-wrapping-without-generating-methods thing, but I haven't checked since.

djberg96:
Replaced by https://github.com/ManageIQ/azure-armrest/pull/79.

PR #22, opened 2015-06-16
VirtualMachineManager updates

djberg96:
This PR revamps the VirtualMachineManager. It updates the list method and adds the list_all method, which will give us all information for all subscription + resource group combinations.

It also no longer assumes a resource group or subscription id is set (and previously it would bomb if they weren't set) so the @base_url instance had to be removed. These must still be optional, so instead I've added a build_url method that instance methods can use to construct url's prior to making requests, and a set_default_subscription private method that will use the first subscription found if not already set.

Lastly, we have to deal with some api_version wonkiness on Microsoft's end, where certain methods currently only work with older api version strings.

Here are some examples:

    vmm = ArmRest::VirtualMachineManager.new

    # List vm's for just one resource group for the current subscription
    p vmm.list('win2k12-test1') 

    # List all vm's for all resource groups for all subscriptions
    p vmm.list_all

### DIFF
--- a/lib/azure/armrest/armrest_manager.rb
+++ b/lib/azure/armrest/armrest_manager.rb
@@ -158,7 +158,7 @@ module Azure
       def resource_groups
         url = url_with_api_version(@base_url, 'subscriptions', subscription_id, 'resourcegroups')
         resp = rest_get(url)
-        JSON.parse(resp.body)
+        JSON.parse(resp.body)["value"]
       end
 
       # Returns information the specified resource group, or the
@@ -208,7 +208,7 @@ module Azure
 
       # Take an array of URI elements and join the together with the API version.
       def url_with_api_version(*paths)
-        File.join(*paths, "?api-version=#{api_version}")
+        path = File.join(*paths) << "?api-version=#{api_version}"
       end
 
     end # ArmRestManager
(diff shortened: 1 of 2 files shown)


bronaghs:
@djberg96 - I like this a lot Daniel. See one or 2 questions, once answered I will merge

lib/azure/armrest/armrest_manager.rb
```diff
@@ -155,10 +155,10 @@ def subscription_info(subscription_id = @subscription_id)
 
       # Returns a list of resource groups for the given subscription.
       #
-      def resource_groups
+      def resource_groups(subscription_id = @subscription_id)
         url = url_with_api_version(@base_url, 'subscriptions', subscription_id, 'resourcegroups')
         resp = rest_get(url)
```
bronaghs: @djberg96 - what is resp is nil?

djberg96: Is that possible? I wouldn't think so.

I did think about wrapping http errors and re-raising them as some sort of armrest error, but that's down the road a bit.

djberg96: Oh, wait, I think you meant empty, not nil. It should just return an empty array, which is fine.

PR #23, opened 2015-06-17
Add thread support for list and list_all in VirtualMachineManager

djberg96:
This threads the list and list_all methods so that multiple requests are made concurrently rather than sequentially.

We will eventually want to do this for any method that makes multiple requests, but I just want to start with these for now.

### DIFF
--- a/lib/azure/armrest/virtual_machine_manager.rb
+++ b/lib/azure/armrest/virtual_machine_manager.rb
@@ -40,15 +40,21 @@ module Azure
           res.empty? ? res : res.map{ |vm| vm['properties'] }
         else
           arr = []
+          thr = []
 
-          resource_groups.each{ |group|
+          resource_groups.each do |group|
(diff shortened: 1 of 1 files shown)


Fryguy:
Be careful here as too many threads simultaneously can cause you to either
run out of file descriptors, or you could get throttled on the other side,
perhaps even banned.

We did something similar for our RHEV integration and ended up using the
parallel gem, which worked awesome.

See
https://github.com/ManageIQ/ovirt/blob/master/lib/ovirt/inventory.rb#L167

Merged #23 <https://github.com/ManageIQ/azure-armrest/pull/23>.

—
Reply to this email directly or view it on GitHub
<https://github.com/ManageIQ/azure-armrest/pull/23#event-333412333>.

djberg96:
@Fryguy Yep, I planned to add a thread capper of some sort at some point before the first release. I didn't know about the parallel gem, which looks very nice, thanks.

PR #25, opened 2015-06-19
Fix for list and list_all methods.

djberg96:
This addresses a bug I introduced when I switched to just retrieving a single element.


Fryguy:
:+1:

PR #26, opened 2015-06-24
Updates to the StorageAccountManager

djberg96:
This PR udpates the get and list methods so they actually work, and I added some private helper functions.

### DIFF
--- a/lib/azure/armrest/storage_account_manager.rb
+++ b/lib/azure/armrest/storage_account_manager.rb
@@ -46,22 +46,22 @@ module Azure
           url = build_url(@subscription_id, group)
           JSON.parse(rest_get(url))['value'].first
         else
-          arr = []
-          thr = []
+          array = []
+          threads = []
 
           resource_groups.each do |group|
             @api_version = '2014-06-01' # Must be set after resource_groups call
             url = build_url(@subscription_id, group['name'])
 
-            thr << Thread.new{
-              res = JSON.parse(rest_get(url))['value'].first
-              arr << res if res
-            }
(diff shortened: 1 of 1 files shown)


lib/azure/armrest/storage_account_manager.rb
```diff
@@ -15,9 +15,54 @@ class StorageAccountManager < ArmRestManager
...
+
+            thr << Thread.new{
+              res = JSON.parse(rest_get(url))['value'].first
+              arr << res if res
+            }
+          end
 
-        @base_url += "resourceGroups/#{@resource_group}/"
-        @base_url += "providers/Microsoft.Storage/storageAccounts"
+          thr.each{ |t| t.join }
```
Fryguy: Prefer `thr.each(&:join)`.  Also why the abbreviations?  I think a variable called `threads` would be much clearer.

lib/azure/armrest/storage_account_manager.rb
```diff
@@ -15,9 +15,54 @@ class StorageAccountManager < ArmRestManager
...
+          JSON.parse(rest_get(url))['value'].first
+        else
+          arr = []
+          thr = []
+
+          resource_groups.each do |group|
+            @api_version = '2014-06-01' # Must be set after resource_groups call
+            url = build_url(@subscription_id, group['name'])
+
+            thr << Thread.new{
```
Fryguy: Minor but prefer do/end for multiline blocks for style.

djberg96:
@Fryguy Ok, I've updated the PR.

PR #27, opened 2015-07-07
Armrest vs ArmRest

djberg96:
I finally got annoyed enough to make this change. Multiple commits caused by annoying differences between Linux and OSX shell commands.


Fryguy:
yay!

PR #28, opened 2015-07-08
Set default subscription in ArmrestManager.configure method

djberg96:
This updates the ArmrestManager.configure method to set a default subscription if one isn't explicitly specified and raises an error if no associated subscriptions are found. I also added some documentation.


Fryguy:
:+1:

djberg96:
@Fryguy or @bronaghs, please merge if it looks alright to you.

PR #32, opened 2015-07-10
Add locations and publishers methods

djberg96:
This adds two new methods, locations and publishers. The locations method returns just a plain array of locations, either on a per-provider basis or the locations for all providers. The publishers method returns a list of publishers (in the form of JSON objects), which can be narrowed by provider and region.

### DIFF
--- a/lib/azure/armrest/armrest_manager.rb
+++ b/lib/azure/armrest/armrest_manager.rb
@@ -262,7 +262,7 @@ module Azure
       #   arm.publishers('eastus', 'Microsoft.ClassicCompute')
       #
       def publishers(region, provider = 'Microsoft.Compute', subscription_id = @subscription_id)
-        @api_version = '2015-06-15'
+        @api_version = '2015-06-15' # Default api-version won't work
 
         url = url_with_api_version(
           @base_url, 'subscriptions', subscription_id, 'providers',
--- a/lib/azure/armrest/armrest_manager.rb
+++ b/lib/azure/armrest/armrest_manager.rb
@@ -254,13 +254,15 @@ module Azure
         array
       end
 
-      # Returns a list of publishers for the given +provider+ and +region+.
+      # Returns an array of publishers for the given +region+ and +provider+,
+      # or "Microsoft.Compute" if no provider is given.
       #
-      # Example:
+      # Examples:
       #
-      #   arm.publishers('Microsoft.Compute', 'eastus').each{ |pub| puts pub['name'] }
+      #   arm.publishers('eastus')
+      #   arm.publishers('eastus', 'Microsoft.ClassicCompute')
       #
-      def publishers(provider, region, subscription_id = @subscription_id)
+      def publishers(region, provider = 'Microsoft.Compute', subscription_id = @subscription_id)
         @api_version = '2015-06-15'
 
         url = url_with_api_version(
@@ -269,7 +271,7 @@ module Azure
         )
 
         response = rest_get(url)
-        JSON.parse(response.body)
+        JSON.parse(response.body).map{ |element| element['name'] }
       end
 
       # Returns a list of subscriptions for the tenant.
--- a/lib/azure/armrest/armrest_manager.rb
+++ b/lib/azure/armrest/armrest_manager.rb
@@ -209,13 +209,68 @@ module Azure
 
       # Returns information about the specific provider +namespace+.
       #
-      def provider_info(namespace)
-        url = url_with_api_version(@base_url, 'providers', namespace)
+      def provider_info(provider)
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/armrest_manager.rb
```diff
@@ -209,14 +209,71 @@ def providers
...
+      # Returns an array of publishers for the given +region+ and +provider+,
+      # or "Microsoft.Compute" if no provider is given.
+      #
+      # Examples:
+      #
+      #   arm.publishers('eastus')
+      #   arm.publishers('eastus', 'Microsoft.ClassicCompute')
+      #
+      def publishers(region, provider = 'Microsoft.Compute', subscription_id = @subscription_id)
+        @api_version = '2015-06-15'
```
blomquisg: Suggest putting this in a const var so it can be easily found in the class and updated later if necessary.  Alternatively, maybe there's a way to inject this into the gem?  Constructor parameter, or env var?

lib/azure/armrest/armrest_manager.rb
```diff
@@ -209,14 +209,71 @@ def providers
...
+      #--
+      # TODO: I think a 7-day cache may be wise here as the overall list of
+      # locations is unlikely to change from day to day.
+      #
+      def locations(provider = nil)
+        array = []
+
+        if provider
+          info = provider_info(provider)
+          info['resourceTypes'].map{ |rt| array << rt['locations'] }
```
blomquisg: This could be:

```ruby
  array = info['resourceTypes'].map{ |rt| rt['locations'] }
```

In the else block below, I suppose you'd have to do:

``` ruby 
  array << info['resourceTypes'].map{ |rt| rt['locations'] }
```

Also, I'll suggest using `collect` instead of `map` only because in the ManageIQ codebase we've normalized on the `collect`, `select`, `inject`, `reject` methods.  Nothing that crucial though...

lib/azure/armrest/armrest_manager.rb
djberg96: Putting what in a const var? The default provider?

lib/azure/armrest/armrest_manager.rb
djberg96: Well, I'll need to create an array object one way or another. I dunno that I can be more efficient since I'm mangling the data after the fact before returning it anyway. But, I can definitely use collect instead of map if that's preferred.

lib/azure/armrest/armrest_manager.rb
blomquisg: No, the `@api_version` magic string :smile: 

In fact, it seems like it would be better to be able to change that without installing a new version of this gem.  But, that *might* have other implications.  Such as, if the `api_version` changes, then that might mean that the implementation of the gem would have to change to accommodate it.  Not sure, but possible?

Nonetheless, I'd like to see that not hidden in a method call inside the API and rather see it exposed somewhere obvious.  Maybe even in the gemspec ... that way, it's obvious from the gemspec what version of the ARM API is supported by this gem.

djberg96:
Alrighty, made some minor updates.

lib/azure/armrest/armrest_manager.rb
blomquisg: So, the main difference is that `collect` (or `map`) is already instantiating a new array and appending to it implicitly.  So, by appending the value to your own array in the block, you're bypassing the advantage of using `collect`.

lib/azure/armrest/armrest_manager.rb
djberg96: Oh, you can set that either in the configure method or on a per object basis either in the constructor or after the fact on an object. You don't want to set that on a per-gem basis because of the unstable nature of the API at the moment because not every version string works for every method (yay, microsoft). In some cases I explicitly set it within a method because I -happen- to know that the default api_version won't work.

blomquisg: Wow, k ... my misunderstanding then.

Might be worth a comment here saying why you're overriding the api_version for this specific method.

Thanks for the clarification.

djberg96: Let me see if I can bump the default api_version to 2015-06-15 first. If that's problematic, then I'll add a comment.

djberg96: Yeah, trying to set it to the latest breaks some (but not all) of the base methods. I realize that setting it like that inside the method feels hacky, but I don't think end users can reliably be expected to know which api-version strings they're supposed to use. If I don't set it explicitly, we'll end up with a bunch of bug reports that end with "set your api-version string properly".

PR #34, opened 2015-07-14
Dispense with subscription_id arguments in methods

djberg96:
Setting a subscription_id within methods was more important before we didn't set a default in the configure method. But now that we do that, specifying a subscription_id for each and every method simply isn't necessary, and it was becoming cumbersome to support it.

In practice I suspect most people will only be dealing with a single subscription. I know we are. And there's a subscription_id accessor that users can set if needed.


bronaghs:
@djberg96 - Any the default subscriber ID is set in the configure method, right?

djberg96:
@bronaghs yes, and an error is raised if one can't be found.

PR #37, opened 2015-07-15
Better api-version handling and other tweaks

djberg96:
The primary change in the PR is that the private methods that build URL strings now accept a local api-version string as the first argument. The problem with setting the api_version instance variable directly is that subsequent method calls could fail if the version wasn't sufficient for that particular method. Still a nuisance, but at least it doesn't break code.

Another change here is the ability to set a provider in the constructor, which lets users set 'Microsoft.Compute' or 'Microsoft.ClassicCompute'. It will use ClassicCompute by default, but at least it's configurable now, whereas it was hardcoded to Classic before.

There was also a minor change where I removed the VALID_VM_SIZES constant, since we can get that information from the series method now.

### DIFF
--- a/lib/azure/armrest/armrest_manager.rb
+++ b/lib/azure/armrest/armrest_manager.rb
@@ -51,6 +51,7 @@ module Azure
       @@content_type = 'application/json'
       @@accept = 'application/json'
       @@token = nil
+      @@providers = {} # Set in constructor
 
       # Set configuration options globally. If set globally you do not need to
       # pass configuration options to individual manager classes.
@@ -80,9 +81,11 @@ module Azure
       # The other options (grant_type, content_type, accept, token, and
       # api_version) should generally NOT be set by you except in specific
       # circumstances.  Setting them explicitly will likely cause breakage.
+      # The api_version will typically be overridden on a per-provider/resource
+      # basis within subclasses anyway.
       #
       # You may need to associate your application with a subscription using
-      # the portal or the New-AzureRoleAssignment powershell command.
+      # the new portal or the New-AzureRoleAssignment powershell command.
       #
       def self.configure(options)
         options.each do |k,v|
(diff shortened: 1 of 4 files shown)


djberg96:
Note that it's debatable what we should set the default provider too. As all of our test VM's are from ClassicCompute, so I used that, but you could make a case either way.

djberg96:
@bzwei Perhaps we could do that, though it would mean setting it up, possibly in the constructor. That would add some overhead, but maybe it's worth it to avoid this api-version pain.

djberg96:
After some struggles, tests are now passing. @bronaghs, @blomquisg and @bzwei please take a look, and merge if you approve.

lib/azure/armrest/armrest_manager.rb
```diff
@@ -167,6 +170,8 @@ def initialize(options = {})
 
         # Base URL used for REST calls. Modify within method calls as needed.
         @base_url = Azure::Armrest::RESOURCE
+
```
bzwei: I looked at the initialize code, seems `@client_id  = @@client_id || options.fetch(:client_id)` should favor `options` over `@@client_id`, same for the rest.

djberg96: @bzwei Possibly, but let's save that for another PR. That's an easy refactoring.

lib/azure/armrest/armrest_manager.rb
```diff
@@ -389,10 +392,32 @@ def rest_delete(url)
       end
 
       # Take an array of URI elements and join the together with the API version.
-      def url_with_api_version(*paths)
+      def url_with_api_version(api_version, *paths)
         File.join(*paths) << "?api-version=#{api_version}"
       end
 
+       # Build a one-time lookup table for each provider & resource. This
+        # lets subclasses set api-version strings properly for each method
```
bzwei: align the comments

lib/azure/armrest/armrest_manager.rb
```diff
@@ -389,10 +392,32 @@ def rest_delete(url)
...
+        # depending on whichever provider they're using.
+        #
+        # e.g. @@providers['Microsoft.Compute']['virtualMachines']['api_version']
+        #
+        # Note that for methods that don't depend on a resource type should use
+        # the @@api_version class variable instead or set it explicitly as needed.
+        #
+      def set_providers_info
+        if @@providers.empty? && @token
```
bzwei: return earlier if condition not met, and keep the main logic in the main body without a conditional block

PR #40, opened 2015-07-22
Add ImageManager class

djberg96:
This PR adds a VirtualMachineImageManager class to handle methods for dealing with VM images. With this class you can do:

```
vmim = VirtualMachineImageManager.new
p vmim.publishers('eastus')
p vmim.offers('eastus', 'Canonical')
p vmim.skus('UbuntuServer', 'eastus', 'Canonical')
p vmim.versions('14.04.2', 'UbuntuServer', 'eastus', 'Canonical')
```

This does cause some breakage, as the publishers method was moved out of the base class and the offers method was moved out of the VirtualMachineManager class. However, I feel this is a better organization conceptually, and it more closely resembles both the Azure CLI and the old ruby sdk for Azure in terms of organization.

I will eventually add a list method that collates this information in the same manner in which the CLI command "azure vm image list" does.

PR #41, opened 2015-07-23
Implement power operations

djberg96:
This PR implements working versions of the start, stop, restart, deallocate, delete and generalize methods.

These are asynchronous methods, so they do not wait for the status to fully change, but the "return!" method you see in the source will raise an error if the initial op call fails.

At the moment it's up to users to check the status of the VM's. In a separate PR I will add a status method that easily lets users get the status of the VM they just changed the power state of, but I want to think about how it should be implemented for a bit.

### DIFF
--- a/lib/azure/armrest/armrest_manager.rb
+++ b/lib/azure/armrest/armrest_manager.rb
@@ -364,18 +364,20 @@ module Azure
         )
       end
 
-      def rest_put(url)
+      def rest_put(url, body = '')
         RestClient.put(
           url,
+          body,
           :accept        => @accept,
           :content_type  => @content_type,
           :authorization => @token,
         )
       end
 
-      def rest_post(url)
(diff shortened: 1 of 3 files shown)


bronaghs:
@djberg96  - can you squash the 3 commits into 1 so it is easier to follow?

bronaghs:
@djberg96 
Although useful to the community,  for our purposes we do not need a method to check the status of the VM after a power op.
Im wondering if the VirtualMachineManager class is getting too big. I am also expanding it by pulling data in from other namespaces, at some point we should think about a reorg.
Otherwise this looks good.

djberg96:
@bronaghs I actually did pull one method out of it and into the VirtualMachineImageManager in https://github.com/ManageIQ/azure-armrest/pull/40 but I agree, we may need a reorg.

PR #42, opened 2015-07-24
Populate VM hash with IP configurations from Compute.Network namespace.

bronaghs:
When we gather information about virtual machines using the *get_vms* method it does not contain public and private IP addresses, which were there in the ClassicCompute namespace.
Instead, we are given a URI which will provide the following IP address information in a RESTful GET:
 - private IP Address details
-  another URI to the public IP Address

This PR gathers the IP configuration information described above and inserts it into the VM hash returned by the get_vms/list method.

### DIFF
--- a/lib/azure/armrest/virtual_machine_manager.rb
+++ b/lib/azure/armrest/virtual_machine_manager.rb
@@ -98,12 +98,12 @@ module Azure
       # p JSON.parse(resp.body)["value"][0]["properties"]["storageProfile"]
       #
       def list(group = nil)
+        array = []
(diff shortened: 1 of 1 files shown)


djberg96:
How many calls to get_nic_profile are being made per VM? I'm just wondering if we should thread that code. I suppose we could save that for a separate PR. Otherwise, looks good to me.

@bzwei, what do you think?

bronaghs:
@djberg96 - once per NIC. So up to maybe 3 or 4 in a typical VM I think.

PR #43, opened 2015-07-24
Fix the get method, add some handy methods

djberg96:
This PR fixes the vmm.get method which was broke, and adds two wrapper methods for convenience - get_model_view and get_instance_view.


bronaghs:
@djberg96 
why does the group default to the vmname?

djberg96:
@bronaghs I did that because that's what the old portal does when you create a new VM.

PR #46, opened 2015-08-13
Introduce TemplateDeploymentManager

bzwei:


djberg96:
@bzwei Overall it looks good. From a style standpoint I wonder if the resource group should be the last argument to each method to be consistent with the other managers. I guess it's not a huge deal with the way params work in Ruby 2.x, though.

I am a bit confused by the URL, as the docs seem to indicate that "/providers/microsoft.resources" is a requisite part of the URL, but you don't use it. Yet, it seems to work fine. Curious.

On a side note, I looked at provider_info for "microsoft.resources" but saw nothing for deployments specifically, so I guess we're sticking with a hard-coded API for now.

djberg96:
FWIW I tried it with "providers/Microsoft.Resources" and it worked. It also worked with 2015-01-01 as the api_version, though 2015-06-01 failed.

Looking at provider_info('microsoft.resources') in the hope of being able to set the api_version dynamically, I see this:

    "subscriptions"
    "subscriptions/providers"
    "subscriptions/operationresults"
    "resourceGroups"
    "subscriptions/locations"
    "subscriptions/tagnames"
    "links"
    "operations"

None of those jump out at me as the one to use to set the api_version, however. Perhaps not coincidentally, the most recent api_version for any of those is 2015-01-01, so maybe any one of them would do.

bzwei:
@djberg96 I agree style consistence is important. I will move resource group to the last argument in the list. I also noticed the docs requires provider in the url, but I found from the azure command line that it was not used. Not sure which way is better, inclusion or exclusion, for future version compatibility. 

For the part to determine api_version, let's leave it hard coded for now until we are definitely sure how to set it dynamically.

PR #50, opened 2015-08-18
Add separate Resource manager classes

djberg96:
This PR puts the various resource related calls into their own classes. They're basically organized the same way that the MS documentation organizes them. It adds the following classes:

ResourceService
ResourceGroupService
ResourceProviderService

Right now some of the methods that these classes are standalone methods in the base Armrest class. My eventual plan is to create a module using these classes that the other managers can then mixin, but I want to save that for another PR.

This does add one external dependency to the gem. It's called "cache_method". I like it because it's the only gem I could find that lets you specify a timeout on a memoized result. It also lets you specify the manner in which the cache is stored, e.g. Redis, if we ever need that. For the ResourceProviderManager especially, often the results of certain calls are not likely to change on a day to day basis, so they're cached for 24 hours by default. This is configurable.

Adding that gem did require a modification to the gemspec, which I took the opportunity to cleanup, as there was some stuff we just don't need.


djberg96:
I should mention that this requires https://github.com/ManageIQ/azure-armrest/pull/48 be merged as well, since it uses a patch request for one of the methods.

lib/azure/armrest/resource_group_service.rb
```diff
@@ -0,0 +1,90 @@
...
+
+        response = rest_get(URI.escape(url))
+
+        JSON.parse(response.body)["value"]
+      end
+
+      # Creates a new +group+ in +location+ for the current subscription.
+      # You may optionally apply +tags+.
+      #
+      def create_resource_group(group, location, tags = nil)
```
bzwei: we should standardize methods in all `service` classes with `list`, `get`, `create`, `delete`, `update`. This method becomes `create`

lib/azure/armrest/resource_group_service.rb
```diff
@@ -0,0 +1,90 @@
...
+        url << "?api-version=#{api_version}"
+
+        response = rest_put(url, body)
+
+        JSON.parse(response.body)
+      end
+
+      # Delete a resource group from the current subscription.
+      #
+      def delete_resource_group(group)
```
bzwei: `delete`

lib/azure/armrest/resource_group_service.rb
```diff
@@ -0,0 +1,90 @@
...
+        url = File.join(Azure::Armrest::COMMON_URI, subscription_id, 'resourcegroups', group)
+        url << "?api-version=#{api_version}"
+
+        response = rest_delete(url)
+        response.return!
+      end
+
+      # Returns information for the given resource group.
+      #
+      def get_resource_group(group)
```
bzwei: `get`

lib/azure/armrest/resource_group_service.rb
```diff
@@ -0,0 +1,90 @@
...
+        url << "?api-version=#{api_version}"
+
+        response = rest_get(url)
+
+        JSON.parse(response.body)
+      end
+
+      # Updates the tags for the given resource group.
+      #
+      def update_resource_group(group, tags)
```
bzwei: `update`

lib/azure/armrest/resource_provider_service.rb
```diff
@@ -0,0 +1,117 @@
...
+
+        JSON.parse(response.body)["value"]
+      end
+
+      cache_method(:list, cache_time)
+
+      # Return information about a specific +namespace+ provider. The results
+      # of this method are cached.
+      #
+      def get_provider(namespace)
```
bzwei: `get`

lib/azure/armrest/resource_provider_service.rb
```diff
@@ -0,0 +1,117 @@
...
+
+        JSON.parse(response.body)
+      end
+
+      cache_method(:get_provider, cache_time)
+
+      # Returns an array of geo-locations for the given +namespace+ provider.
+      # The results of this method are cached.
+      #
+      def geo_locations(namespace)
```
bzwei: Should we add a verb prefix, such as `list_geo_locations` and `list_api_versions` for below? I have `list_deployment_operations` and `get_deployment_operation` in `TemplateDeploymentService`.

lib/azure/armrest/resource_group_service.rb
djberg96: Agreed. Will update.

lib/azure/armrest/resource_provider_service.rb
djberg96: Agreed, let's be consistent.

bzwei:
@djberg96 They look good to me except there are 11 commits now for a single purpose to introduce a new class. Can you rebase them into one commit before we merge?

djberg96:
Too painful to rebase at this point, easier to resubmit the PR. See https://github.com/ManageIQ/azure-armrest/pull/65

PR #52, opened 2015-08-20
Add configured? method to ArmrestManager

bzwei:
Add class method `configured?` because ArmrestManager class variables only needs to be initialized once for all manager instances. This method can be used to avoid multiple initialization.


bzwei:
@djberg96 I have some thoughts about the class variables. Are you online?

bzwei:
After some discussions, we decide to remove all class variables. This PR is no longer valid.

PR #53, opened 2015-08-20
Modify list method

djberg96:
This modifies the list method so that it returns availability sets from all resource groups if no resource group is specified.

### DIFF
--- a/lib/azure/armrest/availability_set_manager.rb
+++ b/lib/azure/armrest/availability_set_manager.rb
@@ -55,10 +55,30 @@ module Azure
       # List availability sets.
       #
       def list(resource_group = @resource_group)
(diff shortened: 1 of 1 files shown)


lib/azure/armrest/availability_set_manager.rb
```diff
@@ -55,10 +55,31 @@ def get(name, resource_group = @resource_group)
...
+        if resource_group
+          url = build_url(resource_group)
+          response = rest_get(url)
+          array << JSON.parse(response.body)['value']
+        else
+          threads = []
+          mutex = Mutex.new
+
+          resource_groups.each do |group|
+            p group['name']
```
bronaghs: @djberg96 can you remove the print statement

djberg96: lol whoops

PR #56, opened 2015-08-21
Require a configuration object to create services

bzwei:
This PR has two commits. The first one is to rename all managers to services.
The second part is to change the way to create service objects.

Before a service object is instantiated, an `ArmrestConfiguration` object should be retrieved through `ArmrestService.configure`. This configuration object can be passed around to create multiple service objects.

Multiple calls to `ArmrestService.configure` will not create redundant tokens or make redundant calls to get subscription IDs because these parameters are cached.

`@@providers` are still a class level hash because this information is not caller dependent. 

However with this implementation done in a short time there are still some limitations because this code may run in the server for a long time. Using class variable is considered evil. 
1. `@@providers` is static. Any change published in Azure will not be detected and updated. This impact may be minimal since Azure does not update the provider information frequently.
2. Caching tokens and subscription ids can build up the memory usage because there is currently no mechanism to reduce the hash size.


djberg96:
On a side note I eventually want to replace the static providers hash with the Resource managers here: https://github.com/ManageIQ/azure-armrest/pull/50. This caches the results for 24 hours instead of permanently.

We might be able to use that same library (cache_method) to timeout the tokens as well.

djberg96:
I did see this warning when I ran the specs:

spec_helper.rb:8: warning: class variable access from toplevel. This was with 2.2.2p95.

djberg96:
I think the only thing I might like to see different is taking :configuration as an option, that way we don't have to change the arity of all the methods.

djberg96:
This needs an adjustment:

https://github.com/bzwei/azure-armrest/blob/armrest_refactor/lib/azure/armrest/armrest_service.rb#L108-L109

I think it should be:

    response = RestClient.get(
      url,
      :content_type  => config.content_type,
      :authorization => @@tokens[key]
    )

djberg96:
This can be added in a separate PR, but I would like to add a bit of validation in the configure method:

    unless configuration.client_id && configuration.client_key
      raise ArgumentError, "client_id and client_key must be specified"
    end

I'm deliberately omitting tenant_id since, in theory, we may someday allow authentication with just user/pass.

djberg96:
Another minor fix for https://github.com/bzwei/azure-armrest/blob/armrest_refactor/lib/azure/armrest/virtual_machine_service.rb#L89: change group['name'] to rg['name']

PR #58, opened 2015-08-21
Mark some internal methods as private.

djberg96:
This makes the add_network_profile, get_nic_profile and add_power_status private methods. I also added a few specs to make sure they're not publicly visible.


djberg96:
Closing, going to redo it.

PR #60, opened 2015-08-26
Mark some internal methods as private, update capture method

djberg96:
This marks a few of our internal methods as private, and updates the capture method.


bzwei:
This is minor. In the future you may want to use a git pull/rebase command instead of merge, so that no more merge branch commits shown in the history.

djberg96:
@bzwei Yep, I'm not sure how to fix this one, though.

bzwei:
We can leave this one as is.

bzwei:
:+1: LGTM

PR #63, opened 2015-08-28
Storage account revamp

djberg96:
This PR replaces the old code for storage accounts which was basically alpha code that doesn't actually work. This updates the code to incorporate Bill's refactor, and actually works.

### DIFF
--- a/lib/azure/armrest/storage_account_service.rb
+++ b/lib/azure/armrest/storage_account_service.rb
@@ -11,13 +11,11 @@ module Azure
         Standard_RAGRS
       ]
 
-      # Creates and returns a new StorageAccountService (SAM) instance.
+      # Creates and returns a new StorageAccountService (SAS) instance.
       #
       def initialize(_armrest_configuration, options = {})
         super
-
         @provider = options[:provider] || 'Microsoft.Storage'
-
         set_service_api_version(options, 'storageAccounts')
       end
 
@@ -33,7 +31,6 @@ module Azure
         raise ArgumentError, "must specify resource group" unless group
 
         url = build_url(group, account_name)
-
         JSON.parse(rest_get(url))
       end
 
@@ -50,13 +47,12 @@ module Azure
           mutex = Mutex.new
 
           resource_groups.each do |rg|
-            url = build_url(rg['name'])
-
             threads << Thread.new do
-              result = JSON.parse(rest_get(url))['value'].first
+              url = build_url(rg['name'])
+              result = JSON.parse(rest_get(url))['value']
               mutex.synchronize{
                 if result
-                  result['resourceGroup'] = rg['name']
+                  result.each{ |hash| hash['resourceGroup'] = rg['name'] }
                   array << result
                 end
               }
@@ -65,7 +61,7 @@ module Azure
 
           threads.each(&:join)
 
-          array
+          array.flatten
         end
       end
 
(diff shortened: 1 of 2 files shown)


djberg96:
response.return! is handy wrapper provided by rest-client that automatically handles redirects and whatnot based on the return code:

https://github.com/rest-client/rest-client/blob/master/lib/restclient/abstract_response.rb#L69-L85

lib/azure/armrest/storage_account_service.rb
```diff
@@ -23,52 +25,50 @@ def initialize(_armrest_configuration, _options = {})
...
       # Returns a list of available storage accounts for the given subscription
       # for the provided +group+, or all resource groups if none is provided.
       #
-      def list(group = armrest_configuration.resource_group)
+      def list(group = nil)
         if group
-          @api_version = '2014-06-01'
-          url = build_url(armrest_configuration.subscription_id, group)
+          url = build_url(group)
           JSON.parse(rest_get(url))['value'].first
```
bzwei: shouldn't it be `JSON.parse(rest_get(url))['value']` without `.first`?

lib/azure/armrest/storage_account_service.rb
```diff
@@ -76,79 +76,120 @@ def list(group = armrest_configuration.resource_group)
...
+        body = {
+          :name       => name,
+          :location   => location,
+          :tags       => tags,
+          :properties => properties
+        }.to_json
 
-        url
+        response = rest_put(url, body)
+        response.return!
```
bzwei: Should we parse the result `JSON.parse(response.return!)` instead of returning the raw return? I believe in a normal case the return is the created storage account. What's the return look like if the creating fails?

lib/azure/armrest/storage_account_service.rb
```diff
@@ -76,79 +76,120 @@ def list(group = armrest_configuration.resource_group)
...
-        url
-      end
-
+      #
+      def delete(account_name, group = armrest_configuration.resource_group)
+        raise ArgumentError, "must specify resource group" unless group
 
+        url = build_url(group, account_name)
+        response = rest_delete(url)
+        response.return!
```
bzwei: Same above, `JSON.parse(response.return!)`?

lib/azure/armrest/storage_account_service.rb
djberg96: More often than not there's nothing to parse. I think it bombs out in that case, but I can double check.

lib/azure/armrest/storage_account_service.rb
djberg96: Yes. I need to look at the other service classes because I don't think we actually need an array, either.

lib/azure/armrest/storage_account_service.rb
djberg96: Yep, if you try to call JSON.parse on a response that doesn't actually return anything, you're going to get "A JSON text must at least contain two octets! (JSON::ParserError)".

bzwei: Why do we want to take `response.return!` as the return object? We should either return a created account object, or simply return `nil`. More importantly, how do the caller know the creation succeeded or failed? Can we analyze the response and convert it properly?

djberg96: Basically the user isn't expecting a return object for any non-get type requests because almost none of the underlying PUT/DELETE/POST calls return a body. The caller knows creation succeeded because it didn't raise an error.

bzwei:
As discussed, we will handle the return type for `create` and `delete` in another PR. This one looks good now. :+1:

PR #66, opened 2015-09-03
Catch exception from RestClient and convert to ArmrestException

bzwei:


djberg96:
Looks good to me. I know this means we'll need some code and spec updates for MIQ, though. If Bronagh's fine with it, I'm fine with it.

djberg96:
Hm, does this need to be rebased? Strange errors.

PR #69, opened 2015-09-15
Fix bug in Resource#list method

djberg96:
This updates/fixes the list method, as it needs to use the internal configuration object and instance variable rather than old instance methods.


bzwei:
Can you double check line 51: `def move(source_group, source_subscription = @subscription_id)`? The default value should be `armrest_configuration.subscription_id`

djberg96:
Good catch. Updated.

PR #71, opened 2015-09-16
Fix list and provider= methods, and other tweaks

djberg96:
This fixes and updates some things in the VirtualMachineService class.

* Fixes the provider= method. This is currently broke.
* The add_network_profile and add_power_status methods blow up when a specific resource group is provided, so they are now only called if no resource group is provided. More on that below.
* We never really went back and updated the way the default resource groups were being set after Bill's configuration PR. This addresses that.
* I commented out the create and update methods, since they aren't actually implemented yet.
* The code now explicitly raises an error in most methods if a resource group is not provided.
* I refactored the list method slightly, only creating a mutex if there's a result. This matches what we do in the other service class list methods.
* I flip-flopped the arguments to the capture method to be in line with other service classes, and now allow the resource_group to be specified as an option.
* I added some more basic specs, and commented out the create spec for now.

Regarding the add_network_profile and add_power_status methods, I feel like this might be a bit of MIQ code needs bleeding into our gem, and wonder if they wouldn't be better suited in the Azure::VM class there instead. But, we can discuss that separately.

### DIFF
--- a/lib/azure/armrest/virtual_machine_service.rb
+++ b/lib/azure/armrest/virtual_machine_service.rb
@@ -18,9 +18,7 @@ module Azure
       #
       def initialize(_armrest_configuration, options = {})
         super
-
         @provider = options[:provider] || 'Microsoft.Compute'
-
         set_service_api_version(options, 'virtualMachines')
       end
 
@@ -28,7 +26,7 @@ module Azure
       # the api_version used for future requests. In practice, only
       # 'Microsoft.Compute' or 'Microsoft.ClassicCompute' should be used.
       #
-      def provider=(name)
+      def provider=(name, options = {})
         @provider = name
         set_service_api_version(options, 'virtualMachines')
       end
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -116,15 +113,20 @@ def list(group = nil)
...
       #
       # Note that this is a long running operation. You are expected to
       # poll on the client side to check the status of the operation.
       #
-      def capture(vmname, group = vmname, options = {})
+      def capture(vmname, options = {}, group = armrest_configuration.resource_group)
         prefix = options.fetch(:prefix)
         container = options.fetch(:container)
         overwrite = options[:overwrite] || false
+        group ||= options[:resource_group]
```
bzwei: which group to use? I think group other then `armrest_configuration.resource_group` is the first order, then from `options`, finally, `armrest_configuration.resource_group`.

So, let's set `group = nil` in the argument list, and set it to `armrest_configuration.resource_group` in the body as the last resort.

djberg96:
@bzwei Agreed, updated.

PR #72, opened 2015-09-16
fixed #70: provider name is now case insensitive

### LINKED ISSUE #70: When determining api versions for providers, case should be ignored
I ran into an issue where I had set my provider to "Microsoft.Insights" instead of "microsoft.insights". Because it wasn't an exact match it returned the default api version, which it turns out didn't work.

I think the set_service_api_version and/or set_providers_info methods should be adjusted so that they ignore case, since MS ignores the provider's case as well.

bzwei:
Now the key in the hash is lower case.

Plus some minor refactoring.

Rename `@@providers` to `@@providers_hash` to avoid some naming confusion because the class has anther `providers` method.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -53,7 +53,7 @@ module Azure
       # Base url used for REST calls.
       attr_accessor :base_url
 
-      @@providers = {} # Set in constructor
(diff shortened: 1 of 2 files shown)


djberg96:
Looks good to me. I think eventually we'll want to leverage the Resource, ResourceGroup and ResourceProvider service classes.

PR #74, opened 2015-09-18
Update get method, fix api-version

djberg96:
It seems the storage account api is in a bit of flux at the moment, as the api version that we gather dynamically is not actually supported. For now we have to set it to '2015-05-01-preview' as per the docs to get things to work.

I suspect this is tied to a change in the results that are returned by the list_account_keys method. Previously this method returned the keys plus the account info. It seems that now it only returns the key info.

So, I've updated the "get" method to accept an optional 3rd argument that will automatically gather the key information as part of the output if desired. I'm pretty sure we're going to need these keys to get at the metrics table info.

I also added a "list_subscription" method.

### DIFF
--- a/lib/azure/armrest/storage_account_service.rb
+++ b/lib/azure/armrest/storage_account_service.rb
@@ -16,22 +16,29 @@ module Azure
       def initialize(_armrest_configuration, options = {})
         super
         @provider = options[:provider] || 'Microsoft.Storage'
-        set_service_api_version(options, 'storageAccounts')
+        #set_service_api_version(options, 'storageAccounts')
+        @api_version = '2015-05-01-preview' # Must hard code for now
       end
 
       # Return information for the given storage account name for the
       # provided +group+. If no group is specified, it will use the
       # group set in the constructor.
       #
+      # If the +include_keys+ option is set to true, then the keys for that
+      # storage account will be included in the output as well.
+      #
       # Example:
       #
       #   sas.get('portalvhds1234', 'Default-Storage-CentralUS')
       #
-      def get(account_name, group = armrest_configuration.resource_group)
+      def get(account_name, group = armrest_configuration.resource_group, include_keys = false)
         raise ArgumentError, "must specify resource group" unless group
 
         url = build_url(group, account_name)
-        JSON.parse(rest_get(url))
+        results = JSON.parse(rest_get(url))
+        results['properties'].merge!(list_account_keys(account_name, group)) if include_keys
+
+        results
       end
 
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/storage_account_service.rb
```diff
@@ -16,22 +16,29 @@ class StorageAccountService < ArmrestService
       def initialize(_armrest_configuration, options = {})
         super
         @provider = options[:provider] || 'Microsoft.Storage'
-        set_service_api_version(options, 'storageAccounts')
+        #set_service_api_version(options, 'storageAccounts')
+        @api_version = '2015-05-01-preview' # Must hard code for now
```
bronaghs: @djberg96 can we remove the commented out line.

djberg96: I left it in to indicate that we will eventually want to return to that once things stabilize on the MS side.

lib/azure/armrest/storage_account_service.rb
```diff
@@ -16,22 +16,29 @@ class StorageAccountService < ArmrestService
...
+      def get(account_name, group = armrest_configuration.resource_group, include_keys = false)
         raise ArgumentError, "must specify resource group" unless group
 
         url = build_url(group, account_name)
-        JSON.parse(rest_get(url))
+        results = JSON.parse(rest_get(url))
+        results['properties'].merge!(list_account_keys(account_name, group)) if include_keys
+
+        results
       end
```
bronaghs: so we have the option to merge the account keys into the GET method results. What prompted that? We didn't care about the keys before when the account information was returned along with it. Is it that we will need the keys to probe for the metrics? Or have the keys been removed from the results of the GET?

djberg96: The list_account_keys method used to return the same information as get, but with the key information as well. But, something changed on Microsoft's side fairly recently, and now it only returns the keys without the rest of the account info.

I'm pretty sure we'll need the keys to get at the metrics data. Keep in mind that it's optional, and false by default.

lib/azure/armrest/storage_account_service.rb
```diff
@@ -66,6 +73,16 @@ def list(group = nil)
         end
       end
 
+      # List all storage accounts for the current subscription. This does not
+      # include storage account key information.
+      #
+      def list_subscription
```
bzwei: Normally we have pattern `list_###s` to list something in plural. Can we somehow rename this method to reduce misleading? Something like `list_for_current_subscription`, or `list_for_subscription` but I would expect a subscription to be in the argument list.

PR #77, opened 2015-09-23
Add a NetworkSecurityGroupService class.

djberg96:
This is an initial implementation of a wrapper class for gathering NetworkSecurityGroup info. I haven't added create or delete methods yet, but figured we probably won't need that for some time.

### DIFF
--- a/lib/azure/armrest/network/ip_address_service.rb
+++ b/lib/azure/armrest/network/ip_address_service.rb
@@ -1,0 +1,102 @@
+module Azure
+  module Armrest
+    # Class for managing public IP addresss.
+    class IpAddressService < ArmrestService
+
+      # Creates and returns a new IpAddressService instance.
+      #
+      def initialize(_armrest_configuration, options = {})
+        super
+        @provider = options[:provider] || 'Microsoft.Network'
+        set_service_api_version(options, 'publicIPAddresses')
+      end
+
+      # Return information for the given IP address name for the
+      # provided +resource_group+. If no group is specified, it will use
+      # the resource group set in the constructor.
+      #
+      # Example:
+      #
+      #   # Where 'your_ip_name' likely corresponds to a VM name.
+      #   ip.get('your_ip_name', 'your_resource_group')
+      #
+      def get(ip_name, resource_group = armrest_configuration.resource_group)
+        raise ArgumentError, "must specify resource group" unless resource_group
+        url = build_url(resource_group, ip_name)
+        JSON.parse(rest_get(url))
+      end
+
+      # Shortcut method that returns just the IP address for the given public
+      # IP address name.
+      #
+      def get_ip(ip_name, resource_group = armrest_configuration.resource_group)
+        get(ip_name, resource_group)['properties']['ipAddress']
+      end
+
+      # Returns a list of available IP addresss for the given subscription
+      # for the provided +group+, or for all resource groups if no group is specified.
+      #
+      def list(group = nil)
+        if group
+          url = build_url(group)
+          JSON.parse(rest_get(url))['value']
+        else
+          array = []
+          threads = []
+          mutex = Mutex.new
+
+          resource_groups.each do |rg|
+            threads << Thread.new(rg['name']) do |group|
+              url = build_url(group)
+              response = rest_get(url)
+              results = JSON.parse(response)['value']
+              if results && !results.empty?
+                mutex.synchronize{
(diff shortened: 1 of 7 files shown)


djberg96:
Also, while I was at it, I added an IpAddressService class with get, list and list_all methods. Also, I added a handy get_ip method.

lib/azure/armrest/network/ip_address_service.rb
```diff
@@ -0,0 +1,102 @@
...
+
+          threads.each(&:join)
+
+          array.flatten
+        end
+      end
+
+      # List all IP addresss for the current subscription.
+      #
+      def list_all_for_subscription
```
bzwei: What's the difference? `list_all_for_subscription` does not have resource group; `list(nil)` queries for all resource groups and combine them. Do they produce the same results?

lib/azure/armrest/network/network_security_group_service.rb
```diff
@@ -0,0 +1,95 @@
+module Azure
+  module Armrest
+    # Class for managing network security groups.
+    class NetworkSecurityGroupService < ArmrestService
```
bzwei: Comparing with what we have in manageiq, can we call it just `SecurityGroupService`?

lib/azure/armrest/network/ip_address_service.rb
djberg96: As far as I can tell, yes. They seem to have added support for per-subscription calls for just about everything now, though it wasn't always so. However, it doesn't explicitly include resource group information, which could be an issue. We'd have to parse it out of an id field. That's something we could add to the base json model class once we modify the code to return them instead of plain hashes.

djberg96: Which, btw, means we could modify the other classes as well, and drop all the threaded code.

lib/azure/armrest/network/network_security_group_service.rb
djberg96: Well, I think we always match the service type in the others. I would like to leave it if that's ok.

lib/azure/armrest/network/ip_address_service.rb
bzwei: I don't get it how you can get the resource group if it not explicitly included in the return from API. Anyway, why do we need `list_all_for_subscription`? We can modify the code in `list` and replace the threaded code with a simple API call as you said (assuming you can solve the problem getting resource group).

Work in the base model to always include resource group name is a good idea. Currently for the list method for all services do not insert resource group if the name is given in argument for `list(group_name)`. I think we should always insert this info, so the caller does not need to worry whether the returns have resource group or not.

lib/azure/armrest/network/network_security_group_service.rb
bzwei: ok

lib/azure/armrest/network/ip_address_service.rb
djberg96: Basically, we have to parse it out of the :id field. So if we have an id field like this:

    id = "/subscriptions/xyz/resourceGroups/some_group/providers/..."

We would parse it out with something like:

    id[/resourceGroups\/(.*?)\//]

bzwei: yes, that will work. Then we don't need `list_all_for_subscription`, just need to refactor `list`.

djberg96: Ok. I want to leave list_all for now to be consistent with other classes. I'd like to do a big refactor where we remove that in a separate PR if that's alright.

bronaghs:
@bzwei - Ready for merge?

bronaghs:
@djberg96 - does this return the same IP addresses as VirtualMachineService::add_network_profile?

lib/azure/armrest/network/network_interface_service.rb
```diff
@@ -0,0 +1,125 @@
+module Azure
+  module Armrest
+    # Class for managing network security groups.
```
bronaghs: is this comment correct?

bronaghs:
@djberg96 
Sounds like Reserved IPs are  Azure's version of floating IPs: https://www.vnext.be/2015/03/10/microsoft-azure-public-ip-and-reserved-ip/

Can you support gathering reserved IP too? Although the reserved IPs apply to the cloud service, not the individual VMs.

PR #79, opened 2015-09-29
Implement JSON base wrapper classes

djberg96:
This replaces https://github.com/ManageIQ/azure-armrest/pull/21

With JSON wrapper classes we can get real classes back instead of plain hashes, and add custom methods to them as we see fit.

This PR does not modify the service classes. They still return hashes for now, but you can still use these wrapper classes by calling .to_json first. Basically, I wanted this first step reviewed before altering other code.

Some sample code:

    vmm = Azure::Armrest::VirtualMachineService.new(conf)
    json = vmm.get('miqazure-win2k8', 'miqazure-dan1', true).to_json
    vm = Azure::Armrest::VirtualMachine.new(json)

    p vm.properties.hardware_profile.vm_size # => "Standard_A1"
    p vm.properties.storage_profile.os_disk.os_type # => "Windows"
    p vm.properties.os_profile.computer_name # => "miqazure-win2k8"

### DIFF
--- a/lib/azure/armrest.rb
+++ b/lib/azure/armrest.rb
@@ -35,3 +35,7 @@ require 'azure/armrest/template_deployment_service'
 require 'azure/armrest/resource_service'
 require 'azure/armrest/resource_group_service'
 require 'azure/armrest/resource_provider_service'
+
+# JSON wrapper classes. The service classes should require their own
+# wrappers from this point on.
+require_relative 'armrest/model/base_model'
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -1,3 +1,5 @@
+require_relative 'model/base_model'
+
 module Azure
   module Armrest
     # Abstract base class for the other service classes.
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -1,0 +1,96 @@
+require 'delegate'
+require 'ostruct'
+
+module Azure
+  module Armrest
+    # Base class for JSON wrapper classes. Each Service class should have
+    # a corresponding class that wraps the JSON it collects, and each of
+    # them should subclass this base class.
+    class BaseModel < Delegator
+      # Access the json instance variable directly.
+      attr_accessor :json
+
+      # Constructs and returns a new JSON wrapper class. Pass in a plain
+      # JSON string and it will automatically give you accessor methods
+      # that make it behave like a typical Ruby object.
+      #
+      # Example:
+      #   class Person < Azure::ArmRest::BaseModel; end
+      #
+      #   json_string = '{"firstname":"jeff", "lastname":"durand",
+      #     "address": { "street":"22 charlotte rd", "zipcode":"01013"}
+      #   }'
+      #
+      #   # Or whatever your subclass happens to be.
+      #   person = Person.new(json_string)
+      #
+      #   # The JSON properties are now available as methods.
+      #   person.firstname        # => 'jeff'
+      #   person.address.zipcode  # => '01013'
+      #
+      #   # Or you can get back the original JSON if necessary.
+      #   person.json # => Returns original JSON
+      #
+      def initialize(json)
+        @json = json
+        @ostruct = JSON.parse(json, object_class: OpenStruct)
+        __setobj__(@ostruct)
+      end
+
+      # Returns the original JSON string passed to the constructor.
+      def to_json
+        @json
+      end
+
+      # Explicitly convert the object to the original JSON string.
+      def to_s
+        @json
+      end
+
+      # Implicitly convert the object to the original JSON string.
+      def to_str
+        @json
+      end
+
+      protected
+
+      # Interface method required to make delegation work. Do
+      # not use this method directly.
+      def __getobj__
+        @ostruct
+      end
+
+      # A custom Delegator interface method that creates snake_case
+      # versions of the camelCase delegate methods.
+      def __setobj__(obj)
+        obj.methods(false).each{ |m|
+          if m.to_s[-1] != '=' # Must deal with nested ostruct's
+            res = obj.send(m)
+            if res.respond_to?(:each)
+              res.each{ |o| __setobj__(o) if o.is_a?(OpenStruct) }
+            else
+              __setobj__(res) if res.is_a?(OpenStruct)
+            end
+          end
+          snake = m.to_s.gsub(/(.)([A-Z])/,'\1_\2').downcase.to_sym
+          obj.instance_eval("alias #{snake} #{m}") unless snake == m
+        }
+      end
+    end
+
+    # Initial class definitions. Reopen these classes as needed.
+
+    class AvailabilitySet < BaseModel; end
+    class Event < BaseModel; end
+    class Resource < BaseModel; end
+    class ResourceGroup < BaseModel; end
+    class ResourceProvider < BaseModel; end
+    class StorageAccount < BaseModel; end
+    class Subnet < BaseModel; end
+    class TemplateDeployment < BaseModel; end
+    class VirtualMachine < BaseModel; end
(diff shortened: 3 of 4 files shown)


lib/azure/armrest/wrapper/base_wrapper.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      # A custom Delegator interface method that creates snake_case
+      # versions of the camelCase delegate methods.
+      def __setobj__(obj)
+        obj.methods(false).each{ |m|
+          if m.to_s[-1] != '=' # Must deal with nested ostruct's
+            res = obj.send(m)
+            if res.respond_to?(:each)
+              res.each{ |o| __setobj__(o) if o.is_a?(OpenStruct) }
+            else
```
bzwei: `elsif` is more readable

djberg96: Sorry, I don't understand what you mean here.

lib/azure/armrest/wrapper/availability_set.rb
```diff
@@ -0,0 +1,8 @@
+require_relative 'base_wrapper'
+
+module Azure
+  module Armrest
+    class AvailabilitySet < Azure::ArmRest::BaseWrapper
```
bzwei: We don't really need to create these many physical files for wrapper classes. Take an example from https://github.com/rest-client/rest-client/blob/master/lib/restclient/exceptions.rb. We can maintain a constant table in the base class and generate wrapper classes by code

djberg96: The idea here was that we could add custom methods to each class as-needed. It's a lot easier to debug and maintain IMHO.

spec/wrappers/event_spec.rb
```diff
@@ -0,0 +1,24 @@
+########################################################################
+# event_spec.rb
+#
+# Test suite for the Azure::ArmRest::Event json wrapper class.
+########################################################################
+require 'spec_helper'
+
+describe "Event" do
```
bzwei: If we use code to generate wrapper classes, no need for so many spec tests.

djberg96: As custom methods get added to each class over time, having the separate spec files will be easier to deal with.

lib/azure/armrest/wrapper/base_wrapper.rb
```diff
@@ -0,0 +1,81 @@
+require 'delegate'
+require 'ostruct'
+
+module Azure
+  module ArmRest
+    # Base class for JSON wrapper classes. Each Service class should have
+    # a corresponding class that wraps the JSON it collects, and each of
+    # them should subclass this base class.
+    class BaseWrapper < Delegator
```
bzwei: Can we rename `wrapper` to `model` for both the directory and the class name? `Fog` also uses the term `model` although it is not an  `ActiveRecord`

djberg96: You want BaseModel or just Model? I guess I worry about name clashes, but it's not a huge deal.

lib/azure/armrest/wrapper/availability_set.rb
bzwei: Initially we use code to generate wrapper classes. For each individual class if we need to add custom method, then it is the time to physically add a new file.

lib/azure/armrest/wrapper/base_wrapper.rb
bzwei: yes, we can use BaseModel. Should not be a problem, right?

lib/azure/armrest/wrapper/base_wrapper.rb
bzwei: `elsif res.is_a?(OpenStruct)`

lib/azure/armrest/wrapper/base_wrapper.rb
```diff
@@ -0,0 +1,81 @@
...
+          if m.to_s[-1] != '=' # Must deal with nested ostruct's
+            res = obj.send(m)
+            if res.respond_to?(:each)
+              res.each{ |o| __setobj__(o) if o.is_a?(OpenStruct) }
+            else
+              __setobj__(res) if res.is_a?(OpenStruct)
+            end
+          end
+          snake = m.to_s.gsub(/(.)([A-Z])/,'\1_\2').downcase.to_sym
+          obj.instance_eval("alias #{snake} #{m}")
```
bzwei: Thought not hurt, but can we add a check, like `unless snake == m` to avoid most repeated alias?

lib/azure/armrest/wrapper/base_wrapper.rb
```diff
@@ -0,0 +1,81 @@
...
+        obj.methods(false).each{ |m|
+          if m.to_s[-1] != '=' # Must deal with nested ostruct's
+            res = obj.send(m)
+            if res.respond_to?(:each)
+              res.each{ |o| __setobj__(o) if o.is_a?(OpenStruct) }
+            else
+              __setobj__(res) if res.is_a?(OpenStruct)
+            end
+          end
+          snake = m.to_s.gsub(/(.)([A-Z])/,'\1_\2').downcase.to_sym
```
bzwei: Do we have enough evidence the resulting `snake` is a legitimate method name?

lib/azure/armrest/resource_provider_service.rb
```diff
@@ -53,7 +54,8 @@ def list
       def get(namespace)
         url = build_url(namespace)
         response = rest_get(url)
-        JSON.parse(response.body)
+        #JSON.parse(response.body)
+        Azure::Armrest::ResourceProvider.new(response.body)
```
bzwei: What's the proper way to convert an array of wrapper objects, for all `list` methods?

djberg96: This shouldn't be in there, sorry. I must have been tinkering and left it in by mistake.

For list methods, we'll need to .map the results.

lib/azure/armrest/wrapper/base_wrapper.rb
djberg96: Yes.

lib/azure/armrest/wrapper/base_wrapper.rb
djberg96: I'm not sure I follow, since the current code deals with lists.

lib/azure/armrest/wrapper/base_wrapper.rb
djberg96: Do you have a counter example?

bzwei: I don't. We can easily break it with space or special characters in the name, but I doubt Azure API will ever use such strange names.

lib/azure/armrest/resource_provider_service.rb
bzwei: If the response JSON string from the API contains an array at root level, how do you handle? You don't want to use `JSON.parse` to get an array of hashes, and use `.map` to convert each back to a JSON string, do you?

lib/azure/armrest/wrapper/base_wrapper.rb
bzwei: Instead of
```
else
  __setobj__(res) if res.is_a?(OpenStruct)
end
```
we can have
```
elsif res.is_a?(OpenStruct)
  __setobj__(res)
end

lib/azure/armrest/wrapper/base_wrapper.rb
djberg96: Ok. If it breaks we can revisit this code, but for now I think it's fine.

lib/azure/armrest/resource_provider_service.rb
djberg96: The plan is to eventually replace the JSON.parse calls with instances of our new wrapper classes. I thought I would make those changes as a separate PR.

PR #80, opened 2015-09-30
Added list_all_for_subscription method.

djberg96:
This adds the list_all_for_subscription which we've got in some of the other service classes.


djberg96:
No longer necessary with the merge of #86.

PR #81, opened 2015-10-02
Modified the StorageAccountService#create method

djberg96:
This puts the create method in line with the create method from other service classes.

### DIFF
--- a/lib/azure/armrest/storage_account_service.rb
+++ b/lib/azure/armrest/storage_account_service.rb
@@ -71,16 +71,18 @@ module Azure
       # Creates a new storage account, or updates an existing account with the
       # specified parameters. The possible parameters are:
       #
-      # - :name
-      #   Required. The name of the storage account within the specified
-      #   resource stack. Must be 3-24 alphanumeric lowercase characters.
+      # Note that the name of the storage account within the specified
+      # must be 3-24 alphanumeric lowercase characters.
+      #
+      # The options available are as follows:
       #
       # - :validating
       #   Optional. Set to 'nameAvailability' to indicate that the account
       #   name must be checked for global availability.
       #
-      # - :type
-      #   The type of storage account. The default is "Standard_GRS".
+      # - :properties
+      #   - :accountType
+      #     The type of storage account, e.g. "Standard_GRS".
       #
       # - :location
       #   Required: One of the Azure geo regions, e.g. 'West US'.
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/storage_account_service.rb
```diff
@@ -95,41 +97,27 @@ def list_all
...
+      #     :location   => "West US",
+      #     :tags       => {:YourCompany => true}
+      #     :properties => {:accountType => "Standard_ZRS"},
       #   )
       #
       # For convenience you may also specify the :resource_group as an option.
       #
-      def create(options = {}, rgroup = armrest_configuration.resource_group)
-        rgroup ||= options[:resource_group]
+      def create(name, options = {}, rgroup = armrest_configuration.resource_group)
```
bzwei: Interesting. I am doing a refactoring that abstracts all similar create/list/list_all/delete/get methods into a base class. When I came to this class I did the exact same refactoring as you did here.

lib/azure/armrest/storage_account_service.rb
```diff
@@ -95,41 +97,27 @@ def list_all
...
+        rgroup = options.delete(:resource_group) || rgroup
```
bzwei: `options` is directly passed to api call, we don't need to provide such connivence to set `resource_group` here. It can always be passed in as the `rgroup` option.

djberg96: With this code you can do:

    sa.create('foo', {:resource_group => 'x', :something => 'y', :something_else => 'z'})

Instead of:

    sa.create('foo', {:something => 'y', :something_else => 'z'}, 'x')

I just thought the first looked a little nicer and was a little more convenient.

lib/azure/armrest/storage_account_service.rb
```diff
@@ -95,41 +97,27 @@ def list_all
...
-          :location   => location,
-          :tags       => tags,
-          :properties => properties
-        }.to_json
+        url << "&validating=" << validating if validating
 
+        body = options.to_json
         response = rest_put(url, body)
         response.return!
       end
```
bzwei: Can you take a look at `list_account_keys` and `regenerate_storage_account_keys` methods in this class? `list_account_keys` seems having a bug as `group` is undefined.
The return `regenerate_storage_account_keys` should not use `#return!`. This method is used when we provide a block to the post method to handle response, but not after the call.
I have generalized the `create` method. If the response is an empty string, then return `nil`, otherwise, return the wrapped model object.

djberg96: You were right about regenerate_storage_account_keys, I've fixed that. I had no problem with list_account_keys.

All of the put/post/delete requests use return! for now.

djberg96:
Going to close this one and just focus on the regenerate method.

PR #82, opened 2015-10-05
Updated SubnetService class, added VirtualNetworkService class

djberg96:
This PR adds the VirtualNetwork class. It also revamps  the SubnetService class and moves it under the network directory. There are also some specs added.

It also fixes a bug in the NetworkInterfaceClass.


lib/azure/armrest/network/network_interface_service.rb
```diff
@@ -2,7 +2,6 @@ module Azure
   module Armrest
     # Class for managing network interfaces.
     class NetworkInterfaceService < ArmrestService
```
bzwei: any reason the namespace does not match the file path?

lib/azure/armrest/network/subnet_service.rb
```diff
@@ -0,0 +1,72 @@
+# Azure namespace
+module Azure
+  # Armrest namespace
+  module Armrest
+    # Base class for managing subnets
+    class SubnetService < VirtualNetworkService
```
bzwei: same question for namespace and file path for all services under network.

djberg96: I guess I should, but the class names are already a bit long. I guess I was just trying to save my fingers a bit...

lib/azure/armrest/network/virtual_network_service.rb
```diff
@@ -0,0 +1,115 @@
...
+        def get(vn_name, resource_group = armrest_configuration.resource_group)
+          raise ArgumentError, "must specify resource group" unless resource_group
+          url = build_url(resource_group, vn_name)
+          JSON.parse(rest_get(url))
+        end
+
+        # Returns a list of available virtual networks in the current subscription
+        # for the provided +resource_group+.
+        #
+        def list(resource_group = armrest_configuration.resource_group)
```
bzwei: Why the implementation for this service is different from others? Not considering to combine all virtual networks for all resource groups?

djberg96: Because I thought we were going to refactor the other classes to remove the threaded code anyway.

bzwei: That's what I thought too.

PR #85, opened 2015-10-08
Add custom resource_group, tags and inspect methods to BaseModel

djberg96:
This adds a custom resource_group, tags and inspect method to the BaseModel wrapper class.

The resource_group parses the group name from the id field.

The tags method was necessary in order to avoid getting back an OpenStruct. For that particular attribute we want a literal hash.

The inspect method shows nicer output, including the current class name, instead of a vanilla OpenStruct.

I also added some specs.

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -7,6 +7,17 @@ module Azure
     # a corresponding class that wraps the JSON it collects, and each of
     # them should subclass this base class.
     class BaseModel < Delegator
+
+      # Declare that a method should return a plain hash instead of an
+      # OpenStruct instance.
+      #--
+      # TODO: Handle nested properties.
+      def self.hash_properties(method)
+        define_method(method){ @ostruct[method].to_h }
+      end
+
+      hash_properties :tags
+
       # Access the json instance variable directly.
       attr_accessor :json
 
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/model/base_model.rb
```diff
@@ -33,10 +33,21 @@ class BaseModel < Delegator
...
         __setobj__(@ostruct)
       end
 
+      # Return the resource group for the current object.
+      def resource_group
+        @resource_group ||= id[/resourceGroups\/(.+?)\//i, 1] rescue nil
+      end
+
+      # Return a hash of tags associated with the resource.
+      def tags
```
bzwei: Are `tags` common from Azure API returns? Can you give me a few examples? I have not used it. 
Can `tags` be `nil` here?

djberg96: You can have tags, which are key/value pairs, on just about any Azure resource, either via the API or the portal. If there are no tags it will return an empty hash instead.

I've attached some sample tags to a few of my resources that you can see, such as the "miqazure-win2k8" VM, which shows "redhat => True".

djberg96: I think a resource can have anywhere from 10 to 100 tags associated with it, depending on the resource. I don't know how common it will be in the wild, and I think it mostly serves as a form of metadata.

bzwei: I ran some tests. `tags` is optional. So you have to deal with the case that `tags` is `nil`.
Besides, let's make this declarative so any child class can easily add attributes that need to be remain in Hash format, for example `parameters` for deployments. I would prefer a syntax like this
`hash_property :tags, :parameters`

djberg96: If tags is nil it will return an empty hash. I have a spec for that.

I like the idea of a hash_property method.

djberg96: After further review, I think this is more difficult to implement than it's worth. For now I think we should just define methods explictly as we need them.

PR #86, opened 2015-10-12
Refactor get and list methods to return JSON wrapper objects

djberg96:
This PR alters the get/list/list_all methods to return our JSON wrapper classes.

It also removes the threaded code, since the API allows us to get all resources for a given subscription. Please update any +list+ calls with +list_all+ if you want all information.

### DIFF
--- a/lib/azure/armrest/template_deployment_service.rb
+++ b/lib/azure/armrest/template_deployment_service.rb
@@ -1,97 +1,108 @@
 module Azure::Armrest
-# Base class for managing templates and deployments
-class TemplateDeploymentService < ArmrestService
+  # Base class for managing templates and deployments
+  class TemplateDeploymentService < ArmrestService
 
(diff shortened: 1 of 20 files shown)


lib/azure/armrest/template_deployment_service.rb
```diff
@@ -1,97 +1,80 @@
...
+    def list_names(resource_group = armrest_configuration.resource_group)
+      list(resource_group).map(&:name)
+    end
 
-  # Get all deployments in a resource group
-  # If the resource group is nil, then return deployments in all groups
-  def list(resource_group = armrest_configuration.resource_group)
-    if resource_group
+    # Get all deployments for the current subscription in a resource group
+    def list(resource_group = armrest_configuration.resource_group)
```
bzwei: Need to add a `list_all` method

PR #89, opened 2015-10-15
Simplify implementation with ResourceGroupBasedService base class(2)

bzwei:
This is a continuation of #88 
Mainly `VirtualMachineService` and `StorageAccountService` are refactored based on `ResourceGroupBasedService`.

It follows the guideline http://talk.manageiq.org/t/armrest-gem-guideline/973


djberg96:
Looks good to me.

PR #91, opened 2015-10-19
Refactor to encapuslate all attributes to be BaseModel or Hash based

bzwei:
With this work all attributes are by default objects instantiated from a subclass of the `BaseModel` unless they are declared to be remain a `Hash`.

The declare is through `attr_hash` which accepts direct attributes in string or symbol or nested attributes in string only format.

All camel_case methods automatically converted by openstruct are removed.

From the user experience point of view, no openstruct object is exposed.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -409,13 +409,13 @@ module Azure
 
         providers.each do |info|
           provider_info = {}
-          info['resourceTypes'].each do |resource|
-            provider_info[resource['resourceType']] = {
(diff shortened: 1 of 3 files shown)


bzwei:
@djberg96 for whatever reason I could not see which part of the code that you commented on. For the question about removing the non snake-case methods, I want to make the model class looks like a real ruby class, not a delegate for the user experience. Camel case method names are not common in ruby classes. Besides, we have a philosophy to provide methods that are really needed and add more only when necessary in the future.

To use `#inspect` helps to recursively inspect nested attributes.

PR #94, opened 2015-10-22
Use custom OpenStruct to override method collisions

djberg96:
In order to override any Object methods in our wrapper classes, we need to do a couple things. First, we must use a custom subclass of OpenStruct that redefines new_ostruct_member so that it doesn't skip methods that are already defined. Second, we need to undefine existing methods within the constructor to override the default Delegator behavior.

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -7,16 +7,18 @@ module Azure
     # a corresponding class that wraps the JSON it collects, and each of
     # them should subclass this base class.
     class BaseModel < Delegator
+      # Initially inherit the exclusion list from parent class or create an empty Set.
       def self.excl_list
-        # initially inherit the exclusion list from parent class or create an empty Set
         @excl_list ||= superclass.respond_to?(:excl_list, true) ? superclass.send(:excl_list) : Set.new
       end
+
       private_class_method :excl_list
 
+      # Merge the declared exclusive attributes to the existing list.
       def self.attr_hash(*attrs)
-        # merge the declared exclusive attributes to the existing list
         @excl_list = excl_list | Set.new(attrs.map(&:to_s))
       end
+
       private_class_method :attr_hash
 
       attr_hash :tags
@@ -62,7 +64,18 @@ module Azure
           @json = json
         end
 
-        @ostruct = OpenStruct.new(hash)
+        ostruct = OpenStruct.new(hash)
+        keys = ostruct.to_h.keys
+
+        # If the method already exists then create an "_alias" method for it.
+        methods.each do |m|
+          if keys.include?(m)
+            ostruct.class.instance_eval{ alias_method "_#{m}", m }
+            instance_eval{ define_singleton_method("_#{m}"){ ostruct[m] } }
+          end
+        end
+
+        @ostruct = ostruct
         super(@ostruct)
       end
 
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -70,7 +70,6 @@ module Azure
         # If the method already exists then create an "_alias" method for it.
         methods.each do |m|
           if keys.include?(m)
-            ostruct.class.instance_eval{ alias_method "_#{m}", m }
             instance_eval{ define_singleton_method("_#{m}"){ ostruct[m] } }
           end
         end
--- a/spec/models/base_model_spec.rb
+++ b/spec/models/base_model_spec.rb
@@ -114,6 +114,16 @@ describe "BaseModel" do
       expect(base.address).to respond_to(:street)
       expect(base.address).to respond_to(:zipcode)
     end
+
+    it "defines an underscore alias for any existing methods" do
+      Object.class_eval{ def temp_stuff; 'hi'; end }
+      json = {:name => 'test', :temp_stuff => 33}.to_json
+      base = Azure::Armrest::BaseModel.new(json)
+
+      expect(base).to respond_to(:_temp_stuff)
+      expect(base._temp_stuff).to eq(33)
+      expect(base.temp_stuff).to eq('hi')
+    end
   end
 
   context "dynamic accessors" do
--- a/spec/models/base_model_spec.rb
+++ b/spec/models/base_model_spec.rb
@@ -117,12 +117,17 @@ describe "BaseModel" do
 
     it "defines an underscore alias for any existing methods" do
       Object.class_eval{ def temp_stuff; 'hi'; end }
-      json = {:name => 'test', :temp_stuff => 33}.to_json
+      Object.class_eval{ def tempStuff; 'hello'; end }
+
+      json = {:name => 'test', :temp_stuff => 33, :tempStuff => 44}.to_json
       base = Azure::Armrest::BaseModel.new(json)
 
       expect(base).to respond_to(:_temp_stuff)
+      expect(base).to respond_to(:_tempStuff)
       expect(base._temp_stuff).to eq(33)
+      expect(base._tempStuff).to eq(44)
       expect(base.temp_stuff).to eq('hi')
+      expect(base.tempStuff).to eq('hello')
     end
   end
 


djberg96:
I've updated the PR so that it will create an _alias for each redefined method. So, if we need to get the original behavior you can use the underscore method. I also added some specs.

bzwei:
I tried another fix (https://gist.github.com/bzwei/ccfbe7a00bc1eea5ad33) by inheriting `BaseModel` from `BasicObject`. `BasicObject` has much fewer methods than `Object` thus has less chance of method name conflicts.

However I don't quite like my fix either. The caller may use some common tools to manage objects, for example, to have a high level error handling mechanism where `Object#inspect` is called to dump the object for logging. These tools rely on `Object`'s methods in the right behaviors without knowing `BaseModel`'s specific methods. The tool will fail If `BasicObject` cannot provide these methods.

For the similar reason the common management tools or generic handling mechanism will fail if we change the expected behaviors of `Object`'s methods or need to use the underscore version. On the other hand the developer (or QA) will test each specific method from `BaseModel` when it is used in the business logic. If a problem is discovered then the underscore method can be tried case by case. Therefore I still prefer to use underscore method for `BaseModel` but not for `Object`

djberg96:
@bzwei Just to be clear, you're saying you want this instead?

    instance_eval{ define_singleton_method("_#{m}"){ ostruct[m] } }

lib/azure/armrest/model/base_model.rb
```diff
@@ -62,7 +74,20 @@ def initialize(json)
           @json = json
         end
 
-        @ostruct = OpenStruct.new(hash)
+        ostruct = AzureOpenStruct.new(hash)
+        keys = ostruct.to_h.keys
+
+        # Override the Delegate's normal behavior so that it any name collisions
+        # become methods for our model rather than skipped. An _alias is created
```
Fryguy: This feels backwards...I would think you should let the original methods work as expected...and `_alias` methods to be the only entry point.  The other option is to just use a BasicObject instead, and then you don't have nearly as many collisions.

djberg96: That approach basically means we should always use hash style syntax within the armrest gem then, because we could never know in advance if an external library that requires it has defined custom Object or BasicObject methods.

bzwei:
@djberg96 Yes. Do we still need `AzureOpenStruct`?

djberg96:
Alrighty, we'll do it your way. I've updated the PR.

lib/azure/armrest/model/base_model.rb
```diff
@@ -62,7 +64,18 @@ def initialize(json)
...
         end
 
-        @ostruct = OpenStruct.new(hash)
+        ostruct = OpenStruct.new(hash)
+        keys = ostruct.to_h.keys
+
+        # If the method already exists then create an "_alias" method for it.
+        methods.each do |m|
+          if keys.include?(m)
+            ostruct.class.instance_eval{ alias_method "_#{m}", m }
```
bzwei: This line is not needed and actually dangerous as it permanently adds a method to `OpenStruct`

lib/azure/armrest/model/base_model.rb
```diff
@@ -62,7 +64,18 @@ def initialize(json)
...
+            instance_eval{ define_singleton_method("_#{m}"){ ostruct[m] } }
```
bzwei: It creates a new method to the hosting object, which is OK, but all our other delegating methods are defined in the delegated object.

My suggestion is continue to define all delegating methods in the delegated object in `__setobj__` method. The implementation is very simple, only two lines change. See my example:

```
      def __setobj__(obj)
        host_methods = methods
        excl_list = self.class.send(:excl_list)
        obj.methods(false).each{ |m|
          if m.to_s[-1] != '=' && !excl_list.include?(m.to_s) # Must deal with nested models
            res = obj.send(m)
            if res.is_a?(Array)
              newval = res.map { |elem| elem.is_a?(Hash) ? @embedModel.new(elem) : elem }
              obj.send("#{m}=", newval)
            elsif res.is_a?(Hash)
              obj.send("#{m}=", @embedModel.new(res))
            end
          end

          snake = m.to_s.gsub(/(.)([A-Z])/,'\1_\2').downcase.to_sym
          snake = snake.to_s.prepend('_').to_sym if host_methods.include?(snake)

          begin
            obj.instance_eval("alias #{snake} #{m}; undef :#{m}") unless snake == m
          rescue SyntaxError
            next
          end
        }
      end
```

This way no need to modify the `initialize` method.

djberg96: Did you try this against the test suite for this PR? I see several failures.

bzwei: Actually my above example has problems too. We should not traverse `obj.methods(false)`, instead we should traverse `obj.to_h.keys` similar to your code.

Your original fix does not handle the case that the key is in camel format such as `toS`.

djberg96: That did not work for me either. I've updated the PR for now to remove the unnecessary instance_eval. We can go from there.

bzwei:
Since we add new methods to the hosting object when there are naming conflicts, why not add all methods directly to the hosting object? This way we completely remove the dependency on `Delegator` and `OpenStruct`. Here is the tested implementation (passing all spec tests)
```
module Azure
  module Armrest
    # Base class for JSON wrapper classes. Each Service class should have
    # a corresponding class that wraps the JSON it collects, and each of
    # them should subclass this base class.
    class BaseModel
      # Initially inherit the exclusion list from parent class or create an empty Set.
      def self.excl_list
        @excl_list ||= superclass.respond_to?(:excl_list, true) ? superclass.send(:excl_list) : Set.new
      end

      private_class_method :excl_list

      # Merge the declared exclusive attributes to the existing list.
      def self.attr_hash(*attrs)
        @excl_list = excl_list | Set.new(attrs.map(&:to_s))
      end

      private_class_method :attr_hash

      attr_hash :tags

      # Constructs and returns a new JSON wrapper class. Pass in a plain
      # JSON string and it will automatically give you accessor methods
      # that make it behave like a typical Ruby object. You may also pass
      # in a hash.
      #
      # Example:
      #   class Person < Azure::ArmRest::BaseModel; end
      #
      #   json_string = '{"firstname":"jeff", "lastname":"durand",
      #     "address": { "street":"22 charlotte rd", "zipcode":"01013"}
      #   }'
      #
      #   # Or whatever your subclass happens to be.
      #   person = Person.new(json_string)
      #
      #   # The JSON properties are now available as methods.
      #   person.firstname        # => 'jeff'
      #   person.address.zipcode  # => '01013'
      #
      #   # Or you can get back the original JSON if necessary.
      #   person.to_json # => Returns original JSON
      #
      def initialize(json)
        # Find the exclusion list for the model of next level (@embedModel)
        # '#' is the separator between levels. Remove attributes
        # before the first separator.
        child_excl_list = self.class.send(:excl_list).map do |e|
          e.index('#') ? e[e.index('#') + 1 .. -1] : ''
        end
        @embedModel = Class.new(BaseModel) do
          attr_hash *child_excl_list
        end

        if json.is_a?(Hash)
          hash = json
          @json = json.to_json
        else
          hash = JSON.parse(json)
          @json = json
        end

        __setobj__(hash)
      end

      def resource_group
        @resource_group ||= id[/resourceGroups\/(.+?)\//i, 1] rescue nil
      end

      def resource_group=(rg)
        @resource_group = rg
      end

      def to_json
        @json
      end

      def to_s
        @json
      end

      def to_str
        @json
      end

      def inspect
        string = "<#{self.class} "
        method_list = methods(false).select{ |m| !m.to_s.include?('=') }
        string << method_list.map{ |m| "#{m}=#{send(m).inspect}" }.join(", ")
        string << ">"
      end

      def ==(other)
        return false unless other.kind_of?(BaseModel)
        __getobj__ == other.__getobj__
      end

      def eql?(other)
        return false unless other.kind_of?(BaseModel)
        __getobj__.eql?(other.__getobj__)
      end

      def [](key)
        __getobj__[key]
      end

      protected

      # Do not use this method directly.
      def __getobj__
        @hashobj
      end

      # Creates snake_case accessor methods for all hash attributes
      # Use _alias if an accessor conflicts with existing methods
      def __setobj__(obj)
        @hashobj = obj
        host_methods = methods
        excl_list = self.class.send(:excl_list)
        obj.each do |key, value|
          snake = key.to_s.gsub(/(.)([A-Z])/,'\1_\2').downcase
          unless excl_list.include?(snake) # Must deal with nested models
            if value.is_a?(Array)
              newval = value.map { |elem| elem.is_a?(Hash) ? @embedModel.new(elem) : elem }
              obj[key] = newval
            elsif value.is_a?(Hash)
              obj[key] = @embedModel.new(value)
            end
          end

          snake.prepend('_') if host_methods.include?(snake.to_sym)
          instance_eval { define_singleton_method(snake) { obj[key] } }
          instance_eval { define_singleton_method("#{snake}=") { |val| obj[key] = val } }
        end
      end
    end

    # Initial class definitions. Reopen these classes as needed.

    class AvailabilitySet < BaseModel; end
    class Event < BaseModel; end
    class Resource < BaseModel; end
    class ResourceGroup < BaseModel; end
    class ResourceProvider < BaseModel; end
    class StorageAccount < BaseModel; end
    class Subscription < BaseModel; end
    class Tag < BaseModel; end
    class TemplateDeployment < BaseModel
      attr_hash 'properties#parameters', 'properties#outputs'
    end
    class TemplateDeploymentOperation < TemplateDeployment; end
    class Tenant < BaseModel; end
    class VirtualMachine < BaseModel; end
    class VirtualMachineInstance < VirtualMachine; end
    class VirtualMachineModel < VirtualMachine; end
    class VirtualMachineExtension < BaseModel; end
    class VirtualMachineImage < BaseModel; end
    class VirtualMachineSize < BaseModel; end

    module Network
      class IpAddress < BaseModel; end
      class NetworkInterface < BaseModel; end
      class NetworkSecurityGroup < BaseModel; end
      class VirtualNetwork < BaseModel; end
      class Subnet < VirtualNetwork; end
    end
  end
end

require_relative 'storage_account'
```

lib/azure/armrest/model/base_model.rb
```diff
@@ -62,7 +64,17 @@ def initialize(json)
...
         end
 
-        @ostruct = OpenStruct.new(hash)
+        ostruct = OpenStruct.new(hash)
+        keys = ostruct.to_h.keys
+
+        # If the method already exists then create an "_alias" method for it.
+        methods.each do |m|
+          if keys.include?(m)
+            instance_eval{ define_singleton_method("_#{m}"){ ostruct[m] } }
```
bzwei: This does  not solve the problem if `keys` have camel case keys.

djberg96: I added a test just for that. What did I miss?

djberg96:
Closed in favor of #102

PR #97, opened 2015-10-23
Fix all_blobs method

djberg96:
The all_blobs method needs to explicitly pass the key to the containers method.

I also added the :container attribute to the Blob object. This will help comparing blob image names with VM images.

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -49,7 +49,9 @@ module Azure
         doc = Nokogiri::XML(response.body)
 
         doc.xpath('//Blobs/Blob').map do |node|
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/model/storage_account.rb
```diff
@@ -60,7 +62,7 @@ def all_blobs(key = nil)
         array = []
         threads = []
 
-        containers.each do |container|
+        containers(key).each do |container|
           threads << Thread.new(container, key){ |c,k| array << blobs(c.name, k) }
```
bzwei: add a space `|c, k|`

djberg96: Maybe we should add a rubocop file.

PR #101, opened 2015-10-30
Add container_properties and container_acl methods

djberg96:
This PR adds the container_properties and container_acl methods to the StorageAccount class. It also sets a :private? attribute when the container_acl method is called.

These could be used to determine which images to make available for provisioning.

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -7,6 +7,7 @@ module Azure
     class StorageAccount < BaseModel
       # Classes used to wrap container and blob information.
       class Container < BaseModel; end
+      class ContainerProperty < BaseModel; end
       class Blob < BaseModel; end
       class BlobServiceProperty < BaseModel; end
       class BlobServiceStat < BaseModel; end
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/model/storage_account.rb
```diff
@@ -35,6 +36,37 @@ def containers(key = nil)
...
+      # If the returned object does not contain x_ms_blob_public_access then
+      # the container is private to the account owner. You can also use the
+      # :private? method to determine if the account is public or private.
+      #
+      def container_acl(name, key = nil)
+        key ||= properties.key1
+
+        response = blob_response(key, "restype=container&comp=acl", name)
+
+        obj = ContainerProperty.new(response.headers)
```
bzwei: we should set `[:private?]` to the hash before converting it to `ContainerProperty`

djberg96: You mean like this?

    response.headers[:private?] = response.headers.include?(:x_ms_blob_public_access) ? false : true

I can do that, though I'm curious why it matters.

PR #104, opened 2015-11-04
Add the blob_properties method.

djberg96:
This adds the blob_properties method for getting information on a single blob, and the delete_blob method for deleting blobs. I also altered a few methods so that any options are always the last argument.

Among the information the blob_properties method returns is x_ms_copy_source, which could be used to get OS info.

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -9,6 +9,7 @@ module Azure
       class Container < BaseModel; end
       class ContainerProperty < BaseModel; end
       class Blob < BaseModel; end
+      class BlobProperty < BaseModel; end
       class BlobServiceProperty < BaseModel; end
       class BlobServiceStat < BaseModel; end
       class BlobMetadata < BaseModel; end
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/model/storage_account.rb
```diff
@@ -169,6 +185,20 @@ def copy_blob(src_container, src_blob, dst_container, dst_blob = nil, key = nil)
...
+      # Delete the given +blob+ found in +container+.
+      #
+      def delete_blob(container, blob, key = nil, options = {})
+        key ||= properties.key1
+
+        url = File.join(properties.primary_endpoints.blob, container, blob)
+        url += "?snapshot=" + options[:date] if options[:date]
+
+        headers = build_headers(url, key, :verb => 'DELETE')
+        response = RestClient.delete(url, headers)
```
bzwei: Feel strange that we still need to explicitly set `:verb => 'DELETE'` since we are calling `RestClient.delete`

djberg96: That's for azure-signature. It needs the request type in advance to properly generate the signature.

PR #105, opened 2015-11-06
Add a list_all_private_images method

djberg96:
This PR adds the list_all_private images that lets users easily retrieve a list of .vhd uri's that are available for provisioning.

### DIFF
--- a/lib/azure/armrest/storage_account_service.rb
+++ b/lib/azure/armrest/storage_account_service.rb
@@ -70,7 +70,6 @@ module Azure
         end
       end
 
-
       # Returns the primary and secondary access keys for the given
       # storage account as a hash.
       #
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/storage_account_service.rb
```diff
@@ -98,6 +97,36 @@ def regenerate_storage_account_keys(account_name, group = armrest_configuration.
         JSON.parse(response)
       end
 
+      # Returns a list of images that are available for provisioning for all
+      # storage accounts in all resource groups.
+      #
+      def list_all_private_images
+        results = []
+
+        resource_groups.each do |resource_group|
```
bzwei: better multithreaded.

lib/azure/armrest/storage_account_service.rb
```diff
@@ -98,6 +97,36 @@ def regenerate_storage_account_keys(account_name, group = armrest_configuration.
...
+          self.list(resource_group.name).each do |storage_account|
```
bzwei: remove `self` (rubcop complains)

bronaghs:
@djberg96 well we might need a separate method to get uploaded images, in that case this method name would be invalid.

bronaghs:
@djberg96 -
The Azure refresh parser has been designed to gather inventory for a specific region. Can the list_all_private_images method be changed to accept a `location` parameter or a `resource group name` parameter just like we have a "list" method for all the other types.

djberg96:
@bronaghs Alright, I changed the name to just list_private_images and it now takes a resource group.

PR #112, opened 2015-11-17
Added AlertService and EventService

djberg96:
A small step towards the beginning of supporting metrics. This PR adds an AlertService and EventService under the "Insights" namespace for getting Azure alert and event information. Some basic specs are also included.


lib/azure/armrest/insights/alert_service.rb
```diff
@@ -0,0 +1,23 @@
+# Azure namespace
+module Azure
+  # Armrest namespace
+  module Armrest
+    # Insights namesspace
+    module Insights
+      # Base class for managing alert rules.
+      class AlertService < ResourceGroupBasedService
+        # The provider used in requests when gathering Alert information.
+        attr_reader :provider
```
bzwei: We may want to move this to the base service since it appears in multiple classes. Can be done in another PR

PR #113, opened 2015-11-19
Added initial version of tables, table_info and table_data methods

djberg96:
This PR allows us to get at the storage account tables and read them. These tables are what store various bits of metrics data in Azure. You can get a list of the tables with the .tables method, or actually read and filter them with the .table_data method.

I also included some basic specs.

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -15,6 +15,10 @@ module Azure
       class BlobServiceStat < BaseModel; end
       class BlobMetadata < BaseModel; end
 
+      # Classes used to wrap table information
+      class Table < BaseModel; end
+      class TableData < BaseModel; end
+
       # The version string used in headers sent as part any internal http
       # request. The default is 2015-02-21.
       attr_accessor :storage_api_version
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/model/storage_account.rb
```diff
@@ -24,6 +28,52 @@ def initialize(json)
         @storage_api_version = '2015-02-21'
       end
 
+      # Returns a list of tables for the given storage account +key+. Note
+      # that full metadata is returned.
+      #
+      def tables(key = nil)
```
bzwei: with our naming convention, this method better to be `list_tables`

lib/azure/armrest/model/storage_account.rb
```diff
@@ -24,6 +28,52 @@ def initialize(json)
...
+        key ||= properties.key1
+        response = table_response(key, nil, "Tables")
+        JSON.parse(response.body)['value'].map{ |t| Table.new(t) }
+      end
+
+      # Return information about a single table for the given storage
+      # account +key+. If you are looking for the entities within the
+      # table, use the table_data method instead.
+      #
+      def table_info(table, key = nil)
```
bzwei: And this method to be `get_table`

lib/azure/armrest/model/storage_account.rb
```diff
@@ -24,6 +28,52 @@ def initialize(json)
...
+        key ||= properties.key1
+        response = table_response(key, nil, "Tables('#{table}')")
+        Table.new(response.body)
+      end
+
+      # Returns a list of TableData objects for the given table +name+ using
+      # account +key+. The exact nature of the TableData object depends on the
+      # type of table that it is.
+      #
+      def table_data(name, key = nil, options = {})
```
bzwei: `list_table_data`?

djberg96:
@bzwei I didn't follow that convention here because it's not a Service class, and it felt clumsy. I didn't follow that convention for any of the blob methods, and those methods are already being used by MIQ.

I could create aliases, though.

bzwei:
@djberg96 right, I did not realize this is not the service class. Alias methods are not necessary then.

PR #114, opened 2015-12-14
Add RBAC service classes

djberg96:
This adds the Role::Assignment and Role::Definition service classes, along with corresponding model definitions. Some basic specs are included as well.


bzwei:
@djberg96 There are three commits in this PR. Seems only the 3rd is related. Can you rebase and clean up?

djberg96:
@bzwei I don't know why it did that.

PR #115, opened 2015-12-16
[WIP] Configuration overhaul

djberg96:
Do not merge yet. Work in progress.

After hitting an issue with using service classes within other service classes, I decided that it would be better to revamp the configuration handling as a starting point.

Part of that means splitting out the configuration handler into its own class and in its own file. I think this will make maintenance easier than having a struct + block with methods defined on the fly, since it's easier to identify what is happening and where.

Another advantage is that we can more robustly set default values, instead of creating a configuration object, and then setting values after the fact.

TL;DR: I think splitting it out into its own bonafide class will make it easier to maintain, test and develop going forward.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -17,8 +17,6 @@ module Azure
 
       @@providers_hash = {} # Set in constructor
 
-      @@subscriptions = {} # subscription caches
-
       # Create a configuration object based on input options.
       # This object can be used to create service objects.
       #
@@ -57,32 +55,8 @@ module Azure
       # the new portal or the New-AzureRoleAssignment powershell command.
       #
       def self.configure(options)
-        armrest_configuration = Azure::Armrest::Configuration.new(options)
-        armrest_configuration.subscription_id ||= fetch_subscription_id(armrest_configuration)
-
-        armrest_configuration
+        Azure::Armrest::Configuration.new(options)
       end
-
-      def self.fetch_subscription_id(config)
-        return @@subscriptions[config.as_cache_key] if @@subscriptions.has_key?(config.as_cache_key)
-
-        url = File.join(Azure::Armrest::RESOURCE, "subscriptions?api-version=#{config.api_version}")
-
-        response = rest_get(
-          url,
-          :content_type  => config.content_type,
-          :authorization => config.token
-        )
-
-        hash = JSON.parse(response)["value"].first
-
-        raise ArgumentError, "No associated subscription found" if hash.empty?
-
-        id = hash.fetch("subscriptionId")
-        @@subscriptions[config.as_cache_key] = id
-      end
-
-      private_class_method :fetch_subscription_id
 
       # Do not instantiate directly. This is an abstract base class from which
       # all other service classes should subclass, and call super within their
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -4,53 +4,10 @@ module Azure
   module Armrest
     # Abstract base class for the other service classes.
     class ArmrestService
-      ArmrestConfiguration = Struct.new(
-        :client_id,
-        :client_key,
-        :tenant_id,
-        :subscription_id,
-        :resource_group,
-        :api_version,
-        :grant_type,
-        :content_type,
-        :accept,
-        :token,
-        :token_expiration # token expiration local system date
-      ) do
-        @@tokens = Hash.new([])
-
-        def as_cache_key
-          "#{grant_type}_#{tenant_id}_#{client_id}_#{client_key}"
-        end
-
-        def token
-          self[:token], self[:token_expiration] = @@tokens[as_cache_key] if self[:token].nil?
-
-          if self[:token].nil? || Time.now > (self[:token_expiration] || Time.new(0))
-            self[:token], self[:token_expiration] = fetch_token
-          end
-          self[:token]
-        end
-
-        def fetch_token
-          token_url = Azure::Armrest::AUTHORITY + tenant_id + "/oauth2/token"
-
-          response = JSON.parse(ArmrestService.rest_post(
-            token_url,
-            :grant_type    => grant_type,
-            :client_id     => client_id,
-            :client_secret => client_key,
-            :resource      => Azure::Armrest::RESOURCE
-          ))
-          token = 'Bearer ' + response['access_token']
-          @@tokens[as_cache_key] = [token, Time.now + response['expires_in'].to_i]
-        end
-
-        private :fetch_token
-      end
-
       # Configuration to access azure APIs
       attr_accessor :armrest_configuration
+
+      alias configuration armrest_configuration
 
       # Base url used for REST calls.
       attr_accessor :base_url
@@ -59,8 +16,6 @@ module Azure
       attr_accessor :provider
 
       @@providers_hash = {} # Set in constructor
-
-      @@tokens = {} # token caches
 
       @@subscriptions = {} # subscription caches
 
@@ -102,23 +57,10 @@ module Azure
       # the new portal or the New-AzureRoleAssignment powershell command.
       #
       def self.configure(options)
-        configuration = ArmrestConfiguration.new
+        armrest_configuration = Azure::Armrest::Configuration.new(options)
+        armrest_configuration.subscription_id ||= fetch_subscription_id(armrest_configuration)
 
-        options.each do |k,v|
-          configuration[k] = v
-        end
-
-        unless configuration.client_id && configuration.client_key
-          raise ArgumentError, "client_id and client_key must be specified"
-        end
-
-        configuration.api_version     ||= '2015-01-01'
-        configuration.grant_type      ||= 'client_credentials'
-        configuration.content_type    ||= 'application/json'
-        configuration.accept          ||= 'application/json'
-        configuration.subscription_id ||= fetch_subscription_id(configuration)
-
-        configuration
+        armrest_configuration
       end
 
       def self.fetch_subscription_id(config)
@@ -147,7 +89,7 @@ module Azure
       # own constructors.
       #
       def initialize(armrest_configuration, service_name, default_provider, options)
-        self.armrest_configuration = armrest_configuration
+        @armrest_configuration = armrest_configuration
         @service_name = service_name
         @provider = options[:provider] || default_provider
 
--- a/lib/azure/armrest/configuration.rb
+++ b/lib/azure/armrest/configuration.rb
@@ -34,6 +34,9 @@ module Azure
       # Set the time in which the token expires.
       attr_writer :token_expiration
 
+      # Explicitly set the token.
+      attr_writer :token
+
       # Yields a new Azure::Armrest::Configuration objects. Note that you must
       # specify a client_id and client_key. All other parameters are optional.
       #
@@ -83,18 +86,9 @@ module Azure
         @token
       end
 
-      # Set the token value. Setting it in this manner will automatically
-      # set the expiration time for 1 hour. To explicitly set the expiration
-      # time, use the :set_token method instead.
+      # Set the token value and expiration time.
       #
-      def token=(value)
-        set_token(value)
-      end
-
-      # Set the token value and expiration time. If no time is specified, then
-      # the time is set for 1 hour.
-      #
-      def set_token(token, token_expiration = Time.now + 3600)
+      def set_token(token, token_expiration)
         @token, @token_expiration = token, token_expiration
       end
 
--- a/lib/azure/armrest/configuration.rb
+++ b/lib/azure/armrest/configuration.rb
@@ -1,0 +1,128 @@
+module Azure
+  module Armrest
+    class Configuration
+      # Used to store unique token information.
+      @@tokens = Hash.new([])
+
+      # The client ID used to gather token information.
+      attr_accessor :client_id
+
+      # The client key used to gather token information.
+      attr_accessor :client_key
+
+      # The tenant ID used to gather token information.
+      attr_accessor :tenant_id
+
+      # The subscription ID used in requests and to gather token information.
+      attr_accessor :subscription_id
+
+      # The resource group used for a given http request.
+      attr_accessor :resource_group
+
+      # The API version used for a given http request. The default is 2015-01-01.
+      attr_accessor :api_version
+
+      # The grant type. The default is client_credentials
+      attr_accessor :grant_type
+
+      # The content type specified for http requests. The default is 'application/json'
+      attr_accessor :content_type
+
+      # The accept type specified for http request results. The default is 'application/json'
+      attr_accessor :accept
+
+      # Set the time in which the token expires.
+      attr_writer :token_expiration
+
+      # Yields a new Azure::Armrest::Configuration objects. Note that you must
+      # specify a client_id and client_key. All other parameters are optional.
+      #
+      # Example:
+      #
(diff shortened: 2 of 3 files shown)


lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,132 @@
...
+        tokens[key]
+      end
+
+      # Set the value of the tokens class variable hash.
+      def []=(key, value)
+        tokens[key] = value
+      end
+
+      # Returns the tokens class variable hash.
+      def tokens
```
bzwei: We should not expose `tokens` as it is used to manage internal generated tokens (excluding those directly set by user). Same reason `[](key)` and `[]=(key, value)` are not needed.

djberg96: Ok, though we're effectively exposing it by making it a class variable. I was hoping to eventually dispense with that and make it a regular instance variable.

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,132 @@
...
+      attr_accessor :content_type
+
+      # The accept type specified for http request results. The default is 'application/json'
+      attr_accessor :accept
+
+      # The bearer token passed to all http requests.
+      attr_writer :token
+
+      # Set the time in which the token expires.
+      attr_writer :token_expiration
```
bzwei: combine them into one method like `def set_token(token, expiration)` as they need to be set externally at the same time

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,132 @@
...
+      def as_cache_key
+        "#{grant_type}_#{tenant_id}_#{client_id}_#{client_key}"
+      end
+
+      alias cache_key as_cache_key
+
+      # Returns the token for the given cache_key, or sets it if it does not
+      # exist or it has expired.
+      #
+      def token
```
bzwei: This implementation is buggy which causes the spec test failures. 
```
      def token
        @token, @token_expiration = tokens[cache_key] if @token.nil?

        if @token.nil? || Time.now > @token_expiration
          @token, @token_expiration = fetch_token
        end

        @token
      end
```

lib/azure/armrest/configuration.rb
djberg96: Agreed.

lib/azure/armrest/configuration.rb
djberg96: Yep, thanks.

djberg96:
@bzwei For now I've redefined a token= method in order to pass the specs. It will internally call the set_token method with a default token expiration.

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+      # time, use the :set_token method instead.
+      #
+      def token=(value)
+        set_token(value)
+      end
+
+      # Set the token value and expiration time. If no time is specified, then
+      # the time is set for 1 hour.
+      #
+      def set_token(token, token_expiration = Time.now + 3600)
```
bzwei: It is Azure that configures the token expiration. Setting a default expiration has no meaning and misleads.

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+
+        # Set defaults if not already set
+        @accept       ||= 'application/json'
+        @content_type ||= 'application/json'
+        @grant_type   ||= 'client_credentials'
+        @api_version  ||= '2015-01-01'
+      end
+
+      # Returns the tokens class variable hash.
+      def tokens
```
bzwei: We really don't need this method as the hash has no value to user. Exposing it actually incurs dangers of being unintentionally overwritten. Besides this kind of method should be a class method.

djberg96: Agreed, I meant to remove that.

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,122 @@
...
+
+      alias cache_key as_cache_key
+
+      # Returns the token for the given cache_key, or sets it if it does not
+      # exist or it has expired.
+      #
+      def token
+        @token, @token_expiration = @@tokens[cache_key] if @token.nil?
+
+        if @token.nil? || Time.now > (@token_expiration || Time.new(0))
```
bzwei: I am working on another PR where I got a Rubocop warning: Do not use Time.now without zone. Use one of Time.zone.now, Time.now.current, Time.now.in_time_zone, Time.now.utc, Time.now.getlocal, Time.now.iso8601, Time.now.jisx0301, Time.now.rfc3339, Time.now.to_i, Time.now.to_f instead.

I thus recommend to use integer for expiration, thus second part of the condition can be `Time.now.to_i > (@token_expiration || 0)`

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,122 @@
...
+          token_url,
+          :grant_type    => grant_type,
+          :client_id     => client_id,
+          :client_secret => client_key,
+          :resource      => Azure::Armrest::RESOURCE
+        ))
+
+        token = 'Bearer ' + response['access_token']
+
+        @@tokens[cache_key] = [token, Time.now + response['expires_in'].to_i]
```
bzwei: This becomes `[token, Time.now.to_i + response['expiries_in'].to_i]`

lib/azure/armrest/armrest_service.rb
```diff
@@ -102,23 +57,10 @@ def fetch_token
       # the new portal or the New-AzureRoleAssignment powershell command.
       #
       def self.configure(options)
```
bzwei: The old design place the configuration class inside `ArmrestService`, thus make sense to provide this factory method to instantiate a configuration object.

The new design move the configuration class as an independent class. A normal usage will be for a user to directly instantiate a configuration object and pass it to create services. This factory method should remain here only for backward compatible reason. 

However this factory method finds the `subscription_id` if not exist. This makes the factory not obsolete but necessary, which kind of fail the purpose of this refactoring work.

lib/azure/armrest/configuration.rb
djberg96: Are those rails methods? I'm afraid I don't agree with that warning for this project.

lib/azure/armrest/configuration.rb
djberg96: Converting these means we would also have to explicitly type check the :token_expiration argument, too.

lib/azure/armrest/armrest_service.rb
djberg96: I can move that into the configuration class.

lib/azure/armrest/configuration.rb
bzwei: Since expiration can be set from outside, we'd better leave it as type `Date`. However the comparison here should change to `Time.now.utc > (@token_expiration.utc || Time.new(0))`. 

Other improvements can be to initialize `@token_expiration` to `Time.new(0)` in the constructor, and in the setter of `@token_expiration` to convert the input date to correct zone.

djberg96: Ok, I've made your suggested changes.

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,158 @@
...
+      def fetch_token
+        token_url = File.join(Azure::Armrest::AUTHORITY, tenant_id, 'oauth2/token')
+
+        response = JSON.parse(ArmrestService.rest_post(
+          token_url,
+          :grant_type    => grant_type,
+          :client_id     => client_id,
+          :client_secret => client_key,
+          :resource      => Azure::Armrest::RESOURCE
+        ))
```
bzwei: `))`?

djberg96: @bzwei Hm? Are you saying you don't like the formatting? That was just a copy/paste from the source, actually. I can change it, though.

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,158 @@
...
+
+        token = 'Bearer ' + response['access_token']
+
+        @@tokens[cache_key] = [token, Time.now.utc + response['expires_in'].to_i]
+      end
+
+      def fetch_subscription_id
+        return @subscriptions[cache_key] if @subscriptions.has_key?(cache_key)
```
bzwei: Prefer `key?(cache_key)`. Check with @matthewd

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,158 @@
...
+      def set_token(token, token_expiration)
+        @token, @token_expiration = token, token_expiration
+      end
+
+      # Returns the expiration datetime for given key, or the current
+      # cache_key if no key is specified. If the value is nil, return
+      # the epoch time.
+      #
+      def token_expiration(key = nil)
+        key ? @@tokens[key].last : @@tokens[cache_key].last
```
bzwei: how about
```
  key ||= cache_key
  @@tokens[key]
```

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,158 @@
...
+
+      alias cache_key as_cache_key
+
+      # Returns the token for the given cache_key, or sets it if it does not
+      # exist or it has expired.
+      #
+      def token
+        @token, @token_expiration = @@tokens[cache_key] if @token.nil?
+
+        if @token.nil? || Time.now.utc > @token_expiration.utc
```
bzwei: since the setter for `@token_expiration` already forces utc, no need `@token_expiration.utc` here.

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,158 @@
...
+
+        # Set defaults if not already set
+        @accept           ||= 'application/json'
+        @content_type     ||= 'application/json'
+        @grant_type       ||= 'client_credentials'
+        @api_version      ||= '2015-01-01'
+        @token_expiration ||= Time.new(0)
+
+        # Used internally
+        @subscriptions = {}
```
bzwei: I don't see why `@subscriptions` is needed. It stores only one entry. In the old code it is a class variable. 
Consider remove `@subscriptions` or change it to class variable.

lib/azure/armrest/armrest_service.rb
```diff
@@ -102,52 +55,15 @@ def fetch_token
...
-          :authorization => config.token
-        )
-
-        hash = JSON.parse(response)["value"].first
-
-        raise ArgumentError, "No associated subscription found" if hash.empty?
-
-        id = hash.fetch("subscriptionId")
-        @@subscriptions[config.as_cache_key] = id
+        Azure::Armrest::Configuration.new(options)
```
bzwei: Should we document this factory method obsolete and update your wiki to guide users to directly create a configuration object?

lib/azure/armrest/configuration.rb
bzwei: Sorry. I miscounted the paring of parentheses. Never mind.

lib/azure/armrest/configuration.rb
djberg96: Yep. We can set it as the default value in the method, actually.

lib/azure/armrest/configuration.rb
djberg96: In theory you could have multiple subscriptions per configuration.

bzwei: yes. but in your `fetch_subscription_id` only takes the first subscription, and it is referenced by `cache_key` which is unique per configuration.

djberg96: Right, the subscriptions instance var should probably be changed to Hash.new([]). I'm trying to decide how the interface should work.

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,183 @@
+module Azure
+  module Armrest
+    class Configuration
+      # Used to store unique token information.
+      @@tokens = Hash.new([])
```
bzwei: should be `@@tokens = {}`

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,183 @@
...
+      # Example:
+      #
+      #   config = Azure::Armrest::Configuration.new(
+      #     :client_id  => 'xxxx',
+      #     :client_key => 'yyyy',
+      #     :tenant_id  => 'zzzz'
+      #   )
+      #
+      def initialize(hash)
+        @subscriptions = Hash.new([]) # Used internally
```
bzwei: `@subscriptions = {}`

bzwei: should be class variable `@@subscriptions`, referenced by `cache_key`, unique per configuration.

djberg96: I don't agree. Since it's unique per configuration, it does not require a class variable. That's one of the changes I wanted to make.

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,183 @@
...
+
+        url = File.join(Azure::Armrest::RESOURCE, "subscriptions?api-version=#{api_version}")
+
+        response = ArmrestService.rest_get(
+          url,
+          :content_type  => content_type,
+          :authorization => token
+        )
+
+        hash = JSON.parse(response)['value'].first
```
bzwei: Does `JSON.parse(response)['value']` gives you all subscriptions? then all of them should be saved. Otherwise I don't see why we need to maintain a list of subscriptions per configuration.

lib/azure/armrest/configuration.rb
bzwei: The whole implementation you only use one key for the hash, which is `cache_key`.

lib/azure/armrest/configuration.rb
djberg96: Yes, it does.

lib/azure/armrest/configuration.rb
djberg96: Do we ever modify a configuration object after it has been created? If not, I'm wondering why we need a cache_key at all.

bzwei: User can create multiple configuration objects from the same set of ids, but they shall share the same token, thus a key is needed. Alternatively you may want to override `hash` method, but be aware a hashcode is used for more than this purpose.

djberg96: Only because we're using a class variable to store @@tokens, right? This made sense before we explicitly passed configuration objects to Service class initializers, but I don't know how much sense it makes now. Why not just have the Service instance get the token from the configuration object that was passed to it?

Maybe I need some code samples to see the situation you're describing.

bzwei:
```
attrs = {...}
cf1 = Configuration.new(attrs)
cf2 = Configuration.new(attrs)
```
How do we ensure `cf1.token == cf2.token` is `true`? Through `@@tokens[cache_key]`

djberg96: Why do we care if cf1.token == cf2.token in practice, though, if we're passing config objects directly to construct Service objects? Given:

    vms = Azure::Armrest::VirtualMachineService.new(cf1)

The vms object is only going to get its info from cf1, since it has everything it needs there, right?

bzwei: We don't want to create a new token for each duplicated configuration. Just like if you already login to an account in a browser, you can continue to access the account in a new tab without login.

djberg96: I don't know that I agree with that. I think that is something for end users to implement on their end if they want.

But, I'll continue down that path for now. Since you brought up config equality, I'm thinking this class should implement Comparable.

PR #116, opened 2015-12-18
Refactor ArmrestService constructor to setup provider and api_version

bzwei:


bronaghs:
@bzwei 
What is driving these changes?
Looks like the Azure EMS refresh code will need updated to support these changes.

bzwei:
@bronaghs The purpose is to increase code reusability. It should not affect the use in manageiq side. Let me know where it breaks.

bzwei:
@djberg96 Passing `service_name` and `default_provider` to abstract base class is for internal use. Each service class knows its service name and provider. These are mandatory. The `options` hash is set by the caller, giving the user opportunity to overwrite or provide additional properties. This is the general guide, with a couple of exceptions in our implementations.

bronaghs:
@bzwei 
Thanks, I see the value now.

djberg96:
@bzwei Ok, I'm fine with it either way.

PR #118, opened 2016-01-21
Fix series method to work with updated provider code.

djberg96:
This fixes the series (aka sizes) method after the changes from #116.

### DIFF
--- a/lib/azure/armrest/virtual_machine_service.rb
+++ b/lib/azure/armrest/virtual_machine_service.rb
@@ -20,11 +20,11 @@ module Azure
       # as "Basic_A1", though information is included as well.
       #
       def series(location)
-        unless @@providers_hash[provider.downcase] && @@providers_hash[provider.downcase]['locations/vmSizes']
-          raise ArgumentError, "Invalid provider '#{provider}'"
-        end
+        resource = provider_info(provider).resource_types.find{ |rt| rt.resource_type == 'locations/vmSizes' }
 
-        version = @@providers_hash[provider.downcase]['locations/vmSizes']['api_version']
+        raise ArgumentError, "Invalid provider '#{provider}'" unless resource
+
+        version = resource.api_versions.first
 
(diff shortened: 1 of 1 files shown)


lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -20,11 +20,11 @@ def initialize(armrest_configuration, options = {})
       # as "Basic_A1", though information is included as well.
       #
       def series(location)
-        unless @@providers_hash[provider.downcase] && @@providers_hash[provider.downcase]['locations/vmSizes']
```
bzwei: The old code is still valid except the need to change keys to lower case, i.e., `location/vmsizes` because PR #116 normalized all keys to lower case.

lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -20,11 +20,11 @@ def initialize(armrest_configuration, options = {})
...
-          raise ArgumentError, "Invalid provider '#{provider}'"
-        end
+        resource = provider_info(provider).resource_types.find{ |rt| rt.resource_type == 'locations/vmSizes' }
 
-        version = @@providers_hash[provider.downcase]['locations/vmSizes']['api_version']
```
bzwei: same as above. change to `locations/vmsizes`

lib/azure/armrest/virtual_machine_service.rb
djberg96: Ok, but I think using methods is better than accessing the class variable directly anyway.

bzwei: The class variable is a cache. It saves an API call to get the provider info.

djberg96: Hm, I would argue that we should probably cache the providers and provider_info methods then, probably for 24 hours at least. I think that information rarely changes.

PR #119, opened 2016-01-21
Fix case in series method.

djberg96:
This fixes the series method so that it users lowercase, in line with the changes from PR #116.


djberg96:

PR #121, opened 2016-01-28
When fetching default subscriptions, only allow enabled subscriptions

djberg96:
This modifies the fetch_subscription_id singleton method so that it only allows **enabled** subscriptions. A minor update to the spec was also needed.


djberg96:

PR #122, opened 2016-01-29
Only raise a warning if subscription is not enabled.

djberg96:
Upon further review, it seems better to emit a warning, rather than raise an error, if a disabled subscription is found. You can still perform read operations, even with a disabled subscription.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -137,13 +137,20 @@ module Azure
           :authorization => config.token
         )
 
-        subscriptions = JSON.parse(response)["value"]
+        array = JSON.parse(response)["value"]
 
-        hash = subscriptions.find{ |entry| entry['state'] == 'Enabled' }
+        if array.nil? || array.empty?
+          raise ArgumentError, "No associated subscription found"
+        end
 
-        raise ArgumentError, "No enabled or associated subscription found" if hash.nil?
+        # Look for the first enabled subscription, otherwise just take the first subscription found.
+        hash = array.find{ |h| h['state'] == 'Enabled' } || array.first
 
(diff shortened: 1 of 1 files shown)


djberg96:

lib/azure/armrest/armrest_service.rb
```diff
@@ -137,11 +137,15 @@ def self.fetch_subscription_id(config)
           :authorization => config.token
         )
 
-        subscriptions = JSON.parse(response)["value"]
+        hash = JSON.parse(response)["value"].first
```
bzwei: Shouldn't you still need to find the first enabled subscription? Use and warn the first subscription only if no enabled subscription is found. Raise error if no subscription is found at all.

djberg96:
@bzwei Ok, updated the PR to prefer enabled subscriptions first.

lib/azure/armrest/armrest_service.rb
```diff
@@ -137,11 +137,18 @@ def self.fetch_subscription_id(config)
...
+
+        # Look for the first enabled subscription, otherwise just take the first subscription found.
+        hash = array.find{ |h| h['state'] == 'Enabled' } || array.first
+
+        id = hash.fetch('subscriptionId')
 
-        raise ArgumentError, "No enabled or associated subscription found" if hash.nil?
+        warn "Warning: subscription #{id} is not enabled" unless hash['state'] == 'Enabled'
 
         @@subscriptions[config.as_cache_key] = hash.fetch('subscriptionId')
```
bzwei: `@@subscription[config.as_cache_key] = id`

djberg96: Fixed. Forgot to change it back.

djberg96:
@bzwei The rebuild worked this time. :)

PR #123, opened 2016-02-10
New SqlServerService and SqlDatabaseService

bzwei:
Refactor existing code base to extract an abstract class `ResourceGroupBasedSubservice`. It is used for services with two levels in the API path.

Introduce two new services.


djberg96:
Looks good to me. This won't affect any existing code that is using NetworkSecurityGroupService, right?

bzwei:
@djberg96  Committed style fixes. The refactoring does not impact the library usage, including `NetworkSecurityGroupService`

djberg96:
@bzwei Ok, excellent.

PR #124, opened 2016-02-16
Blob's snapshots

AlexLamande:
Hello all !

I'm using azure-armrest and I wanted to upgrade it to manage blob snapshots.
Can you please take a look at my PR and tell me if I'm doing it wrong. I'll fix it asap.

Thanks !

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -14,6 +14,7 @@ module Azure
       class BlobServiceProperty < BaseModel; end
       class BlobServiceStat < BaseModel; end
       class BlobMetadata < BaseModel; end
+      class BlobSnapshot < Blob; end
 
       # Classes used to wrap table information
       class Table < BaseModel; end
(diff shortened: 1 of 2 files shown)


AlexLamande:
Well, I can work on it but as the only revellent thing that is returned is the snap date in headers, I didn't know if returning a ```BlobSnapshot``` made sense.

lib/azure/armrest/model/storage_account.rb
```diff
@@ -249,6 +251,22 @@ def delete_blob(container, blob, key = nil, options = {})
...
+
+        url = File.join(properties.primary_endpoints.blob, container, blob)
+        url += "?comp=snapshot"
+
+        headers = build_headers(url, key, :blob, :verb => 'PUT')
+        headers['Content-Type'] = ''
+        response = RestClient.put(url, '', headers)
+
+        BlobSnapshot.new({
+          :Name => blob,
```
bzwei: Please align by `=>`
Keys here are strings and lower case (unless the raw API returns capitalized). So they are `'name'` and `'snapshot'`

bzwei:
@AlexLamande please rebase all commits into one.

lib/azure/armrest/model/storage_account.rb
```diff
@@ -249,6 +251,22 @@ def delete_blob(container, blob, key = nil, options = {})
...
+
+        url = File.join(properties.primary_endpoints.blob, container, blob)
+        url += "?comp=snapshot"
+
+        headers = build_headers(url, key, :blob, :verb => 'PUT')
+        headers['Content-Type'] = ''
+        response = RestClient.put(url, '', headers)
+
+        BlobSnapshot.new(
+          :name          => blob,
```
bzwei: Keys are string like  `'name' => blob`

AlexLamande: oh ok sorry

bzwei:
LGTM 
@djberg96 please review

djberg96:
Yep, looks good, thanks!

bzwei:
@AlexLamande Thanks.

PR #126, opened 2016-02-24
Add support for proxies, and refactor code to use new generic method.

djberg96:
This adds proxy support to azure-armrest.

This required refactoring the rest_xxx methods internally to use a single generic function called rest_execute, which uses RestClient::Request.execute instead of the rest-client helper methods, which don't support parameters beyond the url and headers.

At the moment, the tests are failing however, and need to be refactored. I am not sure we really want the rest methods exposed as part of the API, either.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -15,7 +15,8 @@ module Azure
         :content_type,
         :accept,
         :token,
-        :token_expiration # token expiration local system date
+        :token_expiration, # token expiration local system date
+        :proxy
       ) do
         @@tokens = Hash.new([])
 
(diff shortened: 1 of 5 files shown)


djberg96:
@bzwei Please take a look when you have the chance.

I'm somewhat baffled how some of the tests are passing currently. For example, this raises a TypeError:

    resp = RestClient.get('')
    
    generic_request.rb:17:in `dup': can't dup NilClass (TypeError)

Which is causing this test (and those like it) to fail:

    expect{ Azure::Armrest::ArmrestService.rest_get(:url => '') }.to raise_error(Azure::Armrest::ApiException)

The question is, how are they passing now?

lib/azure/armrest/armrest_service.rb
```diff
@@ -362,49 +374,64 @@ def self.raise_api_exception(e)
 
       def rest_get(url)
         self.class.rest_get(
-          url,
-          :accept        => armrest_configuration.accept,
-          :content_type  => armrest_configuration.content_type,
-          :authorization => armrest_configuration.token,
+          :url     => url,
```
bzwei: for the instance methods (`rest_get`, `rest_put` etc) they all look very similar. We should be able to extract the composing hash logic into a private helper method

djberg96: Done.

lib/azure/armrest/armrest_service.rb
```diff
@@ -35,13 +36,20 @@ def token
...
           token_url = Azure::Armrest::AUTHORITY + tenant_id + "/oauth2/token"
 
-          response = JSON.parse(ArmrestService.rest_post(
-            token_url,
-            :grant_type    => grant_type,
-            :client_id     => client_id,
-            :client_secret => client_key,
-            :resource      => Azure::Armrest::RESOURCE
-          ))
+          response = JSON.parse(
```
bzwei: why not call `ArmrestService.rest_post` since you already refactored `rest_post` method?

PR #129, opened 2016-02-26
Add armrest_configuration alias

djberg96:
This is largely a cosmetic update. I was just tired of typing out "armest_configuration", so I added a shorter alias and use it internally.

This will not only help save my fingers, it will potentially help avoid Rubocop warnings for lines being too long.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -59,6 +59,8 @@ module Azure
       # Configuration to access azure APIs
       attr_accessor :armrest_configuration
 
+      alias configuration armrest_configuration
+
       # Base url used for REST calls.
       attr_accessor :base_url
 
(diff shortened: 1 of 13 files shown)


lib/azure/armrest/armrest_service.rb
```diff
@@ -230,12 +232,12 @@ def subscriptions
...
       #
-      def subscription_info(subscription_id = armrest_configuration.subscription_id)
+      def subscription_info(subscription_id = configuration.subscription_id)
         url = url_with_api_version(
-          armrest_configuration.api_version,
+          configuration.api_version,
           @base_url,
           'subscriptions',
-          armrest_configuration.subscription_id
+          configuration.subscription_id
```
bzwei: Thank rubocop, here it must be `subcription_id`

djberg96: Yep, good catch by rubocop. There were a couple other minor things it found, and I realized that we had documented and tested one StorageAccount method incorrectly as well.

I pushed up those changes.

PR #130, opened 2016-03-01
Reject api version strings that are ahead of the current date.

djberg96:
Today @bronaghs discovered that MS set an api version string for VM's that's ahead of the current date (2016-03-30), and it's not a preview string either. The problem is that it doesn't work.

So, this patch will reject any api version strings that are ahead of the current date.


bronaghs:
LGTM

PR #134, opened 2016-03-07
Modify the StorageAccountService class so that it sets the proxy on all StorageAccount models

djberg96:
This PR adds the ArmrestService.configuration method. It returns whatever was sent to the ArmrestService.configure method. It also modifies the StorageAccount model's constructor to use the proxy information set in the configure method.

The primary motivation was to carry over the :proxy setting, which is not carried over automatically for the StorageAccount model because it makes its own http requests. However, setting this separately for every StorageAccount model object is cumbersome.

Note that I use a class instance variable, so subclasses will not have this information. It is easy enough to change if we would rather have it as a class variable and allow access for subclasses.

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -246,7 +246,8 @@ module Azure
         response = ArmrestService.rest_put(
           :url     => dst_url,
           :payload => '',
-          :headers => headers
+          :headers => headers,
+          :proxy   => proxy
         )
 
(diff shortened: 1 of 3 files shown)


djberg96:

Also, I just noticed as I was working on this that the :proxy option wasn't being passed to the copy_blob method. That's been updated now, too.

djberg96:
@bzwei Updated as per your suggestions. Please let me know what you think.

PR #136, opened 2016-03-09
attach/detach data disk to vm

sutrix-tai-nguyen:
Hello all !

I'm using azure-armrest to manage azure volumes and I want to add new functions to attach/detach Data Disk (Volume) to one VM base on update VM API.
Can you please take a look at my PR and tell me if there is anything wrong. I'll fix it asap.
If you have better ideas, please let me know.

Thank you very much!

### DIFF
--- a/lib/azure/armrest/virtual_machine_service.rb
+++ b/lib/azure/armrest/virtual_machine_service.rb
@@ -120,11 +120,66 @@ module Azure
         vm_operate('powerOff', vmname, group)
       end
 
+      # Add new data disk to vm
+      # data_disk example
+      #  {
+      #    name:"diskname",
+      #    diskSizeGB:"1",
+      #    lun: 0,
+      #    vhd: {
+      #      uri:"https://myaccount.blob.core.windows.net/vhds/vhdname.vhd"
+      #    },
+      #    createOption: "Empty"
+      #  }
+      def attach_data_disk(vmname, data_disk, rgroup = armrest_configuration.resource_group)
+        vm = get(vmname, rgroup)
+        dds = vm.properties.storage_profile.data_disks
+        if dds.empty?
+          ds = [data_disk]
+        else
+          ds = build_data_disks(dds)
+          ds << data_disk
+        end
+        data = build_data_disks_request(ds, vm.location)
+        update(vmname, rgroup, data)
+      end
+
+      # Remove a data disk from vm data disks list
+      # Require data_disk_name, example: "datadiskname"
+      def detach_data_disk(vmname, data_disk_name, rgroup = armrest_configuration.resource_group)
+        vm = get(vmname, rgroup)
+        dds = vm.properties.storage_profile.data_disks
+        dds.delete_if { |dd| dd.name == data_disk_name }
+        data = build_data_disks_request(build_data_disks(dds), vm.location)
+        update(vmname, rgroup, data)
+      end
+
       def model_class
         VirtualMachineModel
       end
 
       private
+
+      # Build request hash from vm data disk
+      def build_data_disks(data_disks)
+        data_disks.collect do |dd|
+          {
+            :name         => dd.name,
+            :diskSizeGB   => dd.disk_size_gb,
+            :lun          => dd.lun,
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -120,12 +120,102 @@ def stop(vmname, group = armrest_configuration.resource_group)
...
+      #              uri:"https://myaccount.blob.core.windows.net/vhds/vhdname.vhd"
+      #            },
+      #            createOption: "Empty"
+      #          }
+      #        ]
+      #      }
+      #    }
+      #  }
+      #
+      def update(vmname, data, rgroup = armrest_configuration.resource_group)
```
bzwei: The base `ResourceGroupBasedService` already has a `update` method (an alias to `create`), and your implementation is the same. This implementation is thus unnecessary. Besides your return is wrong.

bzwei:
@djberg96 sorry, I did not see you already commented.

sutrix-tai-nguyen:
Hi @bzwei, @djberg96,

Thank you for your comment.
I updated and rebased from upstream.
Please help me review again. Let me know if anything else is still wrong.

Thanks.

lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -120,12 +120,74 @@ def stop(vmname, group = configuration.resource_group)
...
+      end
+
       def model_class
         VirtualMachineModel
       end
 
       private
 
+      # Build request hash from vm data disk
+      def build_data_disks(data_disks)
```
bzwei: You need to resolve all rubocop complains, for example, this method can be rewritten 
```
data_disks.collect do |dd|
  {
    :name         => dd.name,
    :diskSizeGB   => dd.disk_size_gb,
    :lun          => dd.lun,
    :vhd          => {:uri => dd.vhd.uri}
    :createOption => dd.create_option
  }
end 
```

lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -120,12 +120,74 @@ def stop(vmname, group = configuration.resource_group)
...
+      #  {
+      #    name:"diskname",
+      #    diskSizeGB:"1",
+      #    lun: 0,
+      #    vhd: {
+      #      uri:"https://myaccount.blob.core.windows.net/vhds/vhdname.vhd"
+      #    },
+      #    createOption: "Empty"
+      #  }
+      def attach_data_disk(vmname, data_disk, rgroup = armrest_configuration.resource_group)
```
bzwei: Have you tested it with Azure?

sutrix-tai-nguyen: Yes, I have. we can attach Data Disk to VM via update VM.

lib/azure/armrest/virtual_machine_service.rb
sutrix-tai-nguyen: Sorry, I use rubocop and ruby with other version. it used other convention.
Updated, thank you for your comment.

sutrix-tai-nguyen:
Please review again. Thanks

sutrix-tai-nguyen:
Seem all thing is OK. Please re-check and merge. Thank you!

sutrix-tai-nguyen:
Just mixed to 1 commit. Thank you @AlexLamande

djberg96:
@sutrix-tai-nguyen We talked it over some more and we've decided to reject this PR. However, we don't want you to think it's not appreciated!

It's good code, but basically we want to keep these sorts of helper methods out of our code base. Mostly it's a matter of maintenance, and the desire to restrict the methods of this library to just what's absolutely needed. Otherwise, things could get out of hand for us.

I think the best path is if you create your own library to extend this library as you like. Call it "azure-armrest-extensions" or something. That way, you can add whatever convenience methods you want without asking for our permission. It's also likely other people will be happy to contribute to your library with convenient wrappers of their own. We may even suggest some!

Thanks again.

sutrix-tai-nguyen:
@djberg96 I know. No problem. Anyway thank you for your review and comment.

PR #137, opened 2016-03-09
Refactor/virtualmachine object

zhitongLBN:
- Add VirtualMachine, VirtualMachineModel, VirtualMachineInstance, OsDisk, DataDisk object
- VirtualMachineService return these new object in the case of `get`
- Instance of these objects can call methods in services to preform actions

PS:
Hi everyone, I'm new in this project please contact me to help me improve, thx! :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@
 /pkg/
 /spec/reports/
 /tmp/
+tags
+*.sw?
--- a/examples/virtual_machine.rb
+++ b/examples/virtual_machine.rb
@@ -1,0 +1,27 @@
+$LOAD_PATH.unshift File.expand_path('../../lib', __FILE__)
+require_relative '../lib/azure-armrest'
+require 'pp'
+
+
+conf = Azure::Armrest::ArmrestService.configure(
+  :client_id       => 'client_id',
+  :client_key      => 'client_key',
+  :tenant_id       => 'tenant_id',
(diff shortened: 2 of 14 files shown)


bzwei:
@zhitongLBN Thank you for your good effort to help to enhance this gem. We however think this refactoring is too much. The base model is already capable of auto-generating accessor methods. The `os_disk` or `data_disk` of a vm, for example, is already an instance of base model's subclasses, although unnamed. Part of your goal is to manually create named model classes. We acknowledge such needs and will enhance the base class to auto generate named nesting models in a future PR. 

We try not to mingle model and service classes at the gem level. Therefore we cannot adopt your suggestion to include the service in the VM model class. However at your application you can always extend the model class to achieve the same goal. 

Once again we appreciate your help and inspiration.

zhitongLBN:
@bzwei Thank you for your answer, please let me know if you change mind and want some real objects in your gem. I will be grad to help :)

bzwei:
@zhitongLBN I am thinking of declaring the nesting model classes in the parent, for example
```
class OsDisk < BaseModel; end
class DataDisk < BaseModel; end
class VirtualMachineModel < VirtualMachine
  nested_model :osDisk => OsDisk, :dataDisk => DataDisk
end
```
In the parser, i.e., `__setobj__`, if the key matches a declared key, its corresponding model class is used instead of the embed_model.

PR #138, opened 2016-03-10
Added storage methods required for SmartState.

roliveri:
Added StorageAccount#get_blob and support methods,
required to support SmartStat analysis for Azure.


roliveri:
@bronaghs @djberg96 - please review.

PR #139, opened 2016-03-10
Added storage methods required for SmartState.

roliveri:
Added StorageAccount#get_blob and support methods,
required to support SmartStat analysis for Azure.

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -325,7 +325,20 @@ module Azure
 
       # Get the contents of the given +blob+ found in +container+.
       #
-      def get_blob(container, blob, key = nil, options = {})
+      # Low-level method to read a range of bytes from the given +blob+.
+      # Returns the raw http response, giving the caller access to the:
+      # +response.body+::    - the returned data, and the
+      # +response.headers+:: - returned metadata.
+      #
+      # Example:
+      #
+      #   ret = @storage_acct.get_blob(@container, @blob, key, :start_byte => start_byte, :length => length)
+      #   content_md5  = ret.headers[:content_md5].unpack("m0").first.unpack("H*").first
+      #   returned_md5 = Digest::MD5.hexdigest(ret.body)
+      #   raise "Checksum error: #{range_str}, blob: #{@container}/#{@blob}" unless content_md5 == returned_md5
+      #   return ret.body
+      #
+      def get_blob_raw(container, blob, key = nil, options = {})
         key ||= properties.key1
 
         url = File.join(properties.primary_endpoints.blob, container, blob)
@@ -350,6 +363,8 @@ module Azure
         if range_str
           additional_headers['x-ms-range'] = range_str
           additional_headers['x-ms-range-get-content-md5'] = true if options[:md5]
+        else
+          raise ArgumentError, "must specify byte range or entire_image flag" unless options[:entire_image]
         end
 
         headers = build_headers(url, key, :blob, additional_headers)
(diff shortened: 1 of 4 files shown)


roliveri:
@bronaghs @djberg96 - please review.

lib/azure/armrest/model/storage_account.rb
```diff
@@ -323,6 +323,39 @@ def create_blob_snapshot(container, blob, key = nil)
...
+          end
+        end
+
+        if range_str
+          additional_headers['x-ms-range'] = range_str
+          additional_headers['x-ms-range-get-content-md5'] = true if options[:md5]
+        end
+
+        headers = build_headers(url, key, :blob, additional_headers)
+        ArmrestService.rest_get(:url => url, :headers => headers, :proxy => proxy)
```
roliveri: I agree, chose to follow the existing pattern because existing blob_*, table_* and container_* methods should be wrapped in their corresponding models as well. That undertaking should probably be addressed by a separate pull request.

djberg96:
Basically this method is identical to the blob_properties method, except for the range information that you added. The key difference is that blob_properties only returns an object based on the response headers. For your method, you want the response body.

So, you'll basically need to modify the last two lines like so:

    response = ArmrestService.rest_get(:url => url, :headers => headers, :proxy => proxy)
    Blob.new(response.body)

I didn't add this originally because that's a big chunk of data coming over the wire. I have no idea how long that will take. But, there you go.

roliveri:
This method is inherently different from the blob_properties method. While the blob_properties method returns information pertaining to the blob as a whole, blob_get returns information only pertaining to the specific read of data from the blob.

At this level, I don't want only the response body, I want the whole response. This enables higher-level clients to do what they need to do for their specific use case. For example, check the MD5, etc. The clients would then return the body (read data) where appropriate.

Instantiating a blob instance for every blob read doesn't make sense. I think this needs to conform more to a File IO paradigm - once you have a blob instance, you can perform multiple reads from that blob.

djberg96:
@roliveri Yes, I understand. Just return Blob.new(response) instead of the raw response, so that you have a real object to work with.

roliveri:
@djberg96 For my use case, I want the raw response. The SSA code will be performing hundreds, if not thousands of reads - the overhead of instantiating a model wrapper for each read would be unacceptable.

I understand that some clients would benefit from having a blob object returned, but I think that should be implemented by a different method that calls this one.

djberg96:
@roliveri ok, I have a better understanding of the issue now. Bill and I discussed this and I would like two changes.

The first is simple - we need to document that we're returning a raw response rather than on object. I see now why you're doing that, but users need to know. So, please just add some comments, and maybe an example:

    src = "Microsoft.Compute/Images/your_container/some_image.vhd"
    resp = acct.get_blob('system', src, key, :length => 1000)
    resp = acct.get_blob('system', src, key, :range => 100..1000)
    p resp.body

Second, I think we should set a default length so that users don't accidentally try to download the entire blob at once, which would be brutal. Bill and I weren't sure what a good default size was. I was thinking 10mb, but Bill was thinking smaller. Any suggestions?

roliveri:
@djberg96 For your first point, I would take it even further. In addition to doing what you suggest, I think the method should be renamed: `blob_get_raw`. This way, if we want to implement a method that returns the body or blob object, we can name it `blob_get`, and it can call `blob_get_raw`.

As for your second point, I agree, we need a mechanism to prevent users from trying to download the entire image by mistake. Instead of defining a default size however, I'd prefer requiring the user to provide a range. We can define an `:entire_image` option to enable users to download the whole image if desired. So, if the `:entire_image` option isn't set, and no range is provided, the request will fail.

roliveri:
@djberg96 - updated.

PR #140, opened 2016-03-10
Add support for ssl_version and ssl_verify config options

djberg96:
This adds support for two more configuration options - ssl_verify and ssl_version, and passes these values to each http request.

By default the ssl_verify is set to nil, and the ssl_version is set to "TLSv1". As far as I can tell, rest-client is using SSLv3 by default, which is not great. Users can still configure the ssl_version if our default does not suit their needs.

Lastly, this fixes a potential bug in the configure + fetch_subscription_id logic where it wouldn't use the default proxy because it was being called before the default was set. That method should always be called last in the configure method.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -17,7 +17,9 @@ module Azure
         :accept,
         :token,
         :token_expiration, # token expiration local system date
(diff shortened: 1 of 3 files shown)


bronaghs:
@djberg96 - LGTM

PR #141, opened 2016-03-11
Nested model class now is named after the attribute name

bzwei:
The auto-created nested model class used to be anonymous. Now it is named after the attribute name, preserving the path. For example, 
```
irb(main):010:0> v.properties.storage_profile.os_disk.class
Azure::Armrest::VirtualMachineModel::Properties::StorageProfile::OsDisk < Azure::Armrest::BaseModel
``` 

This work was inspired by https://github.com/ManageIQ/azure-armrest/pull/137

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -46,11 +46,8 @@ module Azure
         # Find the exclusion list for the model of next level (@embed_model)
         # '#' is the separator between levels. Remove attributes
         # before the first separator.
-        child_excl_list = self.class.send(:excl_list).map do |e|
+        @child_excl_list = self.class.send(:excl_list).map do |e|
           e.index('#') ? e[e.index('#') + 1..-1] : ''
-        end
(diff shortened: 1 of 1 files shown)


zhitongLBN:
@bzwei 
This will be cool! 
What is more cool is that these objects we retrieved can really use methods.
But I understand if you don't want to maintain your code like this.
PS: 
  also with this approche, some of your object will be repeated like:
   `Azure::Armrest::VirtualMachineModel::Properties::StorageProfile::DataDisk`
you get from vm
and
   `Azure::Armrest::DataDisk`
that you list from disks
and they are the same thing logically

bzwei:
@zhitongLBN Yes we may get two classes that describe the same thing. Ideally it would be better they are one class. There are some reasons that we cannot make them all at the root level like `Azure::Armrest::ModelName`, not to discuss details here.

Luckily Ruby is a scripting language. You can use them interchangeably in most cases unless you want to use `kind_of?` or `is_a?` to force type check.

PR #144, opened 2016-03-15
Add ssl_version and ssl_verify options to the StorageAccount model.

djberg96:
We never added the ssl_version and ssl_verify options to the requests for the StorageAccount model. This addresses that and, like the proxy, uses whatever was set in the Armrest.configure, or TLSv1 for the ssl_version.


bzwei:
@djberg96 It is OK for now. I am a little concerned about directly adding connection parameters as model attributes, but we can consider a refactoring in a future PR.

PR #145, opened 2016-03-16
Fix checking whether nested class already exists

bzwei:
Before we were able to detect naming conflicts for a nested class. As a result a wrong class might be picked to be instantiated for a node.  Now use the fully qualified name of a nested class for testing whether it has been defined.

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -145,7 +145,7 @@ module Azure
       end
 
       def nested_object(klass_name, value)
-        unless self.class.const_defined?(klass_name)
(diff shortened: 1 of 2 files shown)


djberg96:
Weird. JRuby and Ruby 2.0 fail, but the others pass. Should we care?

bzwei:
@djberg96 It bothers me some if Ruby 2.0 failed. Can you help to investigate? I am in the middle of refactoring the base model for the use in another gem. We may consider to port back to here also.

bronaghs:
@djberg96 @bzwei - I agree, I think ruby 2.0 needs to be supported.

Issue #142, opened 2016-03-11
VirtualMachineImageService is broken


bzwei:
The service_name is not set.

PR #146, opened 2016-03-16
Updated VirtualMachineImageService class 

djberg96:
Addresses https://github.com/ManageIQ/azure-armrest/issues/142

### DIFF
--- a/lib/azure/armrest/virtual_machine_image_service.rb
+++ b/lib/azure/armrest/virtual_machine_image_service.rb
@@ -10,26 +10,23 @@ module Azure
       # The publisher used in requests when gathering VM image information.
       attr_accessor :publisher
 
-      # Create and return a new VirtualMachineImageService (VMIM) instance.
+      # Create and return a new VirtualMachineImageService instance.
       #
       # This subclass accepts the additional :location, :provider, and
-      # :publisher options as well. The default provider is set to
-      # 'Microsoft.Compute'.
+      # :publisher options as well.
       #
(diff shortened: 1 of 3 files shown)


lib/azure/armrest/virtual_machine_image_service.rb
```diff
@@ -78,21 +75,21 @@ def skus(offer, location = @location, publisher = @publisher)
       #
       # Example:
       #
-      #   vmim.versions('14.04.2', 'UbuntuServer', 'eastus', 'Canonical')
+      #   vmim.versions('15.10', 'UbuntuServer', 'eastus', 'Canonical').map(&:name)
       #
       #   # sample output
-      #   => ["14.04.201503090", "14.04.201505060", "14.04.201506100", "14.04.201507060"]
+      #   => ["15.10.201511111", "15.10.201511161", "15.10.201512030"]
```
bzwei: The sample output is a string array. The actual return type is an array of `ImageVersion`. What does `ImageVersion` look like?

bzwei:
@djberg96 please address rubocop warnings

lib/azure/armrest/virtual_machine_image_service.rb
djberg96: Note the ".map(&:name) at the end of the example code. :)

The ImageVersion object isn't particularly interesting as it just adds an id field.

djberg96:
@bzwei Addressed the rubocop warnings.

PR #148, opened 2016-03-22
Add logging support.

djberg96:
This adds some basic logging support. It's really just a wrapper for RestClient.log=.

There are also some misc. documentation updates.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -113,21 +113,24 @@ module Azure
       #   - ssl_verify
       #   - ssl_version
       #
-      # Of these, you should include a client_id, client_key and tenant_id.
+      # Of these, you should include a :client_id, :client_key and :tenant_id.
       # The resource_group can be specified here, but many methods allow you
       # to specify a resource group if you prefer flexibility.
       #
-      # If no subscription_id is provided then this method will attempt to find
+      # If no :subscription_id is provided then this method will attempt to find
       # a list of associated subscriptions and use the first one it finds as
       # the default. If no associated subscriptions are found, an ArgumentError
       # is raised.
       #
-      # The other options (grant_type, content_type, accept, token,
-      # token_expirationand api_version) should generally NOT be set by you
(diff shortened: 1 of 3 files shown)


lib/azure/armrest/armrest_service.rb
```diff
@@ -336,6 +346,24 @@ def tenants
...
+      # Sets the log to +output+, which can be a file or a handle.
+      #
+      # Note that because logging is handled via a single class variable
+      # in the underlying RestClient library, you cannot really create
+      # different logs for different service classes. All requests will
+      # go to the same log.
+      #
+      def log=(output)
+        configuration.log = output
+        RestClient.log = output
```
bzwei: Since the `RestClient.log` is a class method, we can also directly add `log` and `log=` as class methods to `ArmrestService`. No need for the additional option in configuration.

bzwei:
When we add support for logging, can we consider the following?
Does the gem need to add its own logging? If yes, write to its own log file or just piggyback to RestClient's log?
Does RestClient open/close the handle for each message? If not, at the application side you need to add logic to prepare the handle and make sure to properly close it on exit. This requires another PR in manageiq side for our own use.

lib/azure/armrest/armrest_service.rb
djberg96: Ok, I'll convert these to class method wrappers and update the specs.

djberg96:
@bzwei In my opinion, we should just piggyback on RestClient's logging, for now at least. While users could set an ENV variable, they might want to explicitly set a log.

I don't think RestClient has any explicit at_exit handling for loggers, but Ruby will automatically cleanup any Ruby objects, including handles, on exit as far as I know. Or the OS will. Anyway, since we're piggy backing on RestClient, IMHO it's the job of RestClient to handle it.

djberg96:
@bzwei I've updated the code. I did make one tweak so that ArmrestService.log returns the actual file name if you assigned a string, instead of the weird abstract class that RestClient generates. Otherwise, it's just a wrapper around RestClient.

PR #149, opened 2016-03-23
Added class method to clear caches.

roliveri:
Added a class method to clear class-level caches in ArmrestService.
This is needed when recording multiple VCR cassettes within a test -
as the tests need to clear the cache between exampes.


roliveri:
@djberg96 @bzwei @bronaghs - please review.

djberg96:
Looks good to me. What do you think @bzwei ?

PR #150, opened 2016-03-24
Capture skip_token when listing events

blomquisg:
* add new `event_list` model that captures the list of the `events` and the `skip_token` if the list of events exceeds the API threshold (which is currently 200 events)
* return `event_list` model from `EventService#list` method
* add new `EventService#list_all` method to return all `events` matching the given `filter` and `select`ion by gathering all the pages of results and concatenating them together
* specs

@djberg96, @bzwei please review
/cc @Fryguy


djberg96:
Nice.

blomquisg:
Thanks @djberg96!

PR #151, opened 2016-03-24
Add MetricsService class

djberg96:
This adds a MetricsService class with a list method for getting a list of metrics definitions.


bronaghs:
@djberg96 
Why get metrics on a resource group basis?, we do not do that for events.

djberg96:
@bronaghs The API requires it:

https://msdn.microsoft.com/en-us/library/azure/dn931939.aspx

I could add a list_all method.

bronaghs:
@djberg96 - yea that would be good.

djberg96:
@bronaghs Actually, I realized once I started that it wasn't practical. The API requires a provider, resource type and name, too. We can add some handy methods in the future as we need them perhaps, but I would like to just start with this.

bronaghs:
OK. Bill, any comments before I merge?

lib/azure/armrest/insights/metrics_service.rb
```diff
@@ -0,0 +1,59 @@
...
+        #   metrics.list('Microsoft.Compute', 'virtualMachines', 'myVm', 'mygroup')
+        #
+        def list(provider, resource_type, resource_name, resource_group = nil, options = {})
+          resource_group ||= configuration.resource_group
+
+          raise ArgumentError, "no resource group provided" unless resource_group
+
+          url = build_url(provider, resource_type, resource_name, resource_group, options)
+
+          response = rest_get(URI.escape(url))
```
bzwei: why do we need `URI.escape` here? should the same be applied for other `rest_get`?

djberg96: Because of the $filter syntax. Here's a URL without escape:

```
"https://management.azure.com/subscriptions/xxx/resourceGroups/yyy/providers/Microsoft.Compute/virtualMachines/zzz/metricDefinitions?api-version=2014-04-01&$filter=name.value eq '\\Processor(_Total)\\% Processor Time'"
```
Same URL now escaped:
```
"https://management.azure.com/subscriptions/xxx/resourceGroups/yyy/providers/Microsoft.Compute/virtualMachines/zzz/metricDefinitions?api-version=2014-04-01&$filter=name.value%20eq%20'%5CProcessor(_Total)%5C%25%20Processor%20Time'"
```
But, this may not be explicitly necessary. It looks to me like rest-client is escaping the parameters automatically, but I can double check. If not, I think we should escape the url's in our core methods.

PR #153, opened 2016-03-24
Deal with all-caps response keys.

djberg96:
It turns out that when getting table data, the response contains duplicate keys, and this causes our auto-method generation to behave strangely. Apparently, the MS solution to get around duplicate keys was to change the case. For example, you will see this in a parsed json response:

```
{"Timestamp@odata.type"=>"Edm.DateTime", "Timestamp"=>"2016-03-24T17:00:07.671801Z", "Average"=>512583270.4, "Count"=>240, "CounterName"=>"\\Memory\\AvailableMemory","Last"=>512753664.0, "Maximum"=>512753664.0,"Minimum"=>511705088.0, "TIMESTAMP@odata.type"=>"Edm.DateTime", "TIMESTAMP"=>"2016-03-24T16:00:00Z","Total"=>123019984896.0}
```
Notice that "timestamp" and "timestamp@odata.type" appears multiple times, differing only in case. The result for an all caps key like this is that it generates "t_im_es_ta_mp" instead of what we expect.

The proposed solution simply checks to see if the key string is all caps and/or contains a non-alpha character. If true, it downcases and capitalizes before proceeding. The result is that we will get "timestamp" and "_timestamp" as expected.

UPDATE: Dispensed with the original plan and switched to using the String#underscore method, and dispensing with our own snake_case method, since we already require active_support.

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -1,3 +1,5 @@
+require 'active_support/core_ext/string/inflections'
+
 module Azure
   module Armrest
     # Base class for JSON wrapper classes. Each Service class should have
@@ -114,7 +116,7 @@ module Azure
         __getobj__[key] = val
 
         return if key_exists
-        add_accessor_methods(snake_case(key), key)
+        add_accessor_methods(key.to_s.underscore, key)
       end
 
       protected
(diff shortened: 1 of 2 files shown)


bzwei:
Sounds to me this is a bug from Microsoft. Can we verify with them? If the timestamp is not useful for now, we may leave it as is.

djberg96:
@bzwei I've sent an email, but have not heard back yet.

djberg96:
From MS: "Timestamp is precise; TIMESTAMP is rounded to the boundary of upload frequency (e.g. if you are tracking in 5-minute intervals, TIMESTAMP will be rounded to the 5-minute boundaries)".

PR #155, opened 2016-03-28
Modify the Insight::Event list method and approach to skip tokens.

djberg96:
Upon further discussion, we decided that we did not like the concept of an EventList class, as it would mean revamping the entire gem to handle skip tokens and lists to match that approach if we wanted to be consistent. Since that would cause serious breakage, this PR takes a different approach.

Within the list method, the last object (only) has the skip_token attribute set. The list_all method then looks for that token on the last object. This required adding the :skip_token attribute to the BaseModel, and modifying the list and list_all methods.

In addition, I added comments to warn users that you can only retrieve events for the last 90 days. Any more than that and you will get an error.

**Update**: we decided to revamp our approach, using a custom collection class (a subclass of Array with a custom attribute) to store the skip token, which the list method returns. We replaced the method prototype for the list method so that it accepts a single hash of options, including an :all option. This obviated the need for a list_all method.

### DIFF
--- a/lib/azure/armrest/insights/event_service.rb
+++ b/lib/azure/armrest/insights/event_service.rb
@@ -13,9 +13,7 @@ module Azure
           super(armrest_configuration, 'eventTypes', 'Microsoft.Insights', options)
         end
 
-        # Returns an +EventList+ containing a list of management evnets for the
-        # current subscription as well as a +next_link+ attribute that can be
-        # used to obtain the next batch of events for this result.
+        # Returns a list of management events for the current subscription.
         #
         # The +filter+ option can be used to filter the result set. Additionally,
         # you may restrict the results to only return certain fields using
@@ -29,7 +27,14 @@ module Azure
         #
         # The +skip_token+ option can be used to grab the next batch of events
         # when the first call reaches the maximum number of events that the API
-        # can return in one batch (API default 200).
+        # can return in one batch (API default 200). You may also set the :all
+        # option to true, in which case all batches will automatically be
+        # collected for you.
+        #
+        # In practice you should always set a filter for eventTimestamp because
+        # you are restricted to 90 days worth of events. If you do not set the
+        # filter and/or you try to retrieve more than 90 days worth of events
+        # then you will get an error. This is a limitation of the Azure API.
         #
         # Example:
         #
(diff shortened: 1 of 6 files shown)


djberg96:
As per our discussion, we actually removed the list_all method, and added an :all option, changing the method prototype to accept a hash of options instead.

lib/azure/armrest/insights/event_service.rb
```diff
@@ -39,63 +44,35 @@ def initialize(armrest_configuration, options = {})
...
-        #
-        def list_all(filter = nil, select = nil)
-          event_list = list(filter, select)
-          events = event_list.events
+          events.skip_token = get_skip_token(json_response['nextLink'])
 
-          while skip_token = event_list.skip_token do
-            event_list = list(filter, select, skip_token)
-            events += event_list.events
+          if options[:all]
```
bzwei: You're mixing the recursive call with looping. The recursive version can be as simple as
```
events.push(*list(options.merge(:skip_token => skip_token)) if options[:all]
```
`skip_token` is local here, you don't need to set it to `events`.

I kind of prefer the looping version, where you should use a while loop, but inside you should update the url with new `skip_token` value and directly call `rest_get`

djberg96: I couldn't get a simple recursive call to work without looping. Otherwise, you get an infinite loop.

lib/azure/armrest/insights/event_service.rb
```diff
@@ -39,63 +44,31 @@ def initialize(armrest_configuration, options = {})
...
-        #   select = "resourceGroupName, operationName"
-        #
-        #   ies.list_all(filter, select).each{ |event|
-        #     p event
-        #   }
-        #
-        def list_all(filter = nil, select = nil)
-          event_list = list(filter, select)
-          events = event_list.events
+          events.skip_token = get_skip_token(json_response['nextLink'])
```
bzwei: you don't need to and should not set `skip_token` to events. Simply `skip_token = get_skip_token(json_response['nextLink'])`

bzwei: Yes, leave this line on

lib/azure/armrest/insights/event_service.rb
```diff
@@ -39,63 +44,31 @@ def initialize(armrest_configuration, options = {})
...
 
-          while skip_token = event_list.skip_token do
-            event_list = list(filter, select, skip_token)
-            events += event_list.events
+          if options[:all] && events.skip_token
+            events.push(*list(options.merge(:skip_token => events.skip_token)))
```
bzwei: clear `events.skip_token` after this line

lib/azure/armrest/model/base_model.rb
```diff
@@ -20,6 +20,9 @@ def self.attr_hash(*attrs)
 
       attr_hash :tags
 
+      # Used for collating partial collections.
```
bzwei: no longer needed in base_model

lib/azure/armrest/model/base_model.rb
```diff
@@ -222,6 +216,10 @@ module Sql
       class SqlServer < BaseModel; end
       class SqlDatabase < BaseModel; end
     end
+
+    class ArmrestCollection < Array
```
bzwei: Even though very simple, I think `ArmrestCollection` should be in its own `.rb` file, under the same directory.

PR #158, opened 2016-04-01
Fix and refactor the StorageAccount#table_data method

djberg96:
This updates the StorageAccount#table_data method, which is currently broken. In addition, it was refactored so that continuation tokens are now supported, and query parsing and string generation is in its own method. I've also updated the documentation.

On a side note, the ArmrestCollection was modified to change the accessor "skip_token" to "continuation_token", because the former implies "$skip" in the query and a String object, which is not necessarily the case. I've updated the EventService class appropriately.

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -65,48 +65,51 @@ module Azure
       # type of table that it is.
       #
       # You may specify :filter, :select or :top as options to restrict your
-      # result set. You may also specify a :nextRowKey, :nextPartitionKey or
-      # :nextTableset for paging.
+      # result set.
       #
       # By default you will receive a maximum of 1000 records. If you wish to
-      # receive more records, you will need to use the continuation token.
+      # receive more records, you will need to use the continuation token. You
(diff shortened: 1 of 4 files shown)


lib/azure/armrest/model/storage_account.rb
```diff
@@ -64,27 +64,55 @@ def table_info(table, key = nil)
...
+      #   # Get the first 10 rows of data from the last 3 days
+      #   date = (Time.now - (86400 * 3)).iso8601
+      #   my_filter = "timestamp ge datetime'#{date}'"
+      #   options = {:top => 10, :filter => my_filter}
+      #
+      #   results = storage_account.table_data(table, key, options)
+      #
+      #   # Now get the next 10 records
+      #   if results.continuation_token
+      #     options.merge!(results.continuation_token)
```
bzwei: should this be `options.merge!(:continuation_token => results.continuation_token)` or simply `options[:continuation_token] = results.continuation_token`?

djberg96: No, because for storage accounts, the continuation_token is actually a hash of up to 3 key/value pairs.

PR #159, opened 2016-04-07
Configuration Overhaul, redux

djberg96:
This moves the configuration options currently being stored within a custom struct into their own class.

This replaces https://github.com/ManageIQ/azure-armrest/pull/115 which had simply gotten too far behind for an easy rebase. As I mentioned before:

After hitting an issue with using service classes within other service classes, I decided that it would be better to revamp the configuration handling as a starting point.

Part of that means splitting out the configuration handler into its own class and in its own file. I think this will make maintenance easier than having a struct + block with methods defined on the fly, since it's easier to identify what is happening and where.

Another advantage is that we can more robustly set default values, instead of creating a configuration object, and then setting values after the fact.

TL;DR: I think splitting it out into its own bonafide class will make it easier to maintain, test and develop going forward.


djberg96:
Ok, I think it's ready to go, rubocop warnings notwithstanding. The highlights are:

* All the configuration stuff has been moved out of the ArmrestService class and into configuration.rb.
* No longer set a default subscription ID. That is now a mandatory argument.
* Backwards compatibility is maintained. Under the hood the ArmrestService.configure method is just calling Configuration.new.
* Class instance variables instead of class variables.
* Updates to the locations, providers and provider_info methods in the ArmrestService class. They now use the configuration data.
* Specs added, updated and improved, which exposed at least one bug.

@bronaghs @bzwei @roliveri @jerryk55 @blomquisg Please review.

lib/azure/armrest/armrest_service.rb
```diff
@@ -242,26 +69,25 @@ def provider_info(provider)
       #--
       #
       def locations(provider = nil)
-        hash = provider.nil? ? @@providers_hash : {provider => @@providers_hash[provider.downcase]}
+        list = configuration.providers
+        list = list.reject { |rp| rp.namespace.casecmp(provider) == 0 } if provider
```
bzwei: should it be `select` rather than `reject`?

lib/azure/armrest/armrest_service.rb
```diff
@@ -242,26 +69,25 @@ def provider_info(provider)
...
+        list.collect { |rp| rp.resource_types.map(&:locations) }.flatten.uniq.sort
       end
 
       # Returns a list of subscriptions for the tenant.
       #
       def subscriptions
         url = url_with_api_version(configuration.api_version, @base_url, 'subscriptions')
         response = rest_get(url)
-        JSON.parse(response.body)["value"].map{ |hash| Azure::Armrest::Subscription.new(hash) }
+        JSON.parse(response.body)['value'].map { |hash| Azure::Armrest::Subscription.new(hash) }
```
bzwei: based on the new naming convention, this method should be `list_subscriptions`

lib/azure/armrest/armrest_service.rb
```diff
@@ -215,23 +42,23 @@ def initialize(armrest_configuration, service_name, default_provider, options)
...
       #
-      def providers
-        url = url_with_api_version(configuration.api_version, @base_url, 'providers')
-        resp = rest_get(url)
-        JSON.parse(resp.body)["value"].map{ |hash| Azure::Armrest::ResourceProvider.new(hash) }
+      def list_providers
+        configuration.providers
       end
 
+      alias providers list_providers
```
bzwei: we used a lot of alias in this library. I am thinking marking these method deprecated if the only purpose is for backward comparability. We may follow the example in http://stackoverflow.com/questions/293981/best-practice-to-mark-deprecated-code-in-ruby

djberg96: Yes, I'm adding aliases for now for backwards compatibility. We can remove them in a future release.

lib/azure/armrest/armrest_service.rb
djberg96: Yes. Will fix.

lib/azure/armrest/armrest_service.rb
```diff
@@ -273,53 +99,32 @@ def subscription_info(subscription_id = configuration.subscription_id)
...
-        url = url_with_api_version(
-          configuration.api_version,
-          @base_url,
-          'subscriptions',
-          configuration.subscription_id,
-          'resourcegroups'
-        )
-        response = rest_get(url)
-        JSON.parse(response)["value"].map{ |hash| Azure::Armrest::ResourceGroup.new(hash) }
+        Azure::Armrest::ResourceGroupService.new(configuration).list
```
bzwei: We talked about making the return an array of names only. Same for `resources`

lib/azure/armrest/armrest_service.rb
```diff
@@ -273,53 +99,32 @@ def subscription_info(subscription_id = configuration.subscription_id)
...
-          resource_group
-        )
-
-        response = rest_get(url)
-        Azure::Armrest::ResourceGroup.new(response.body)
+      def get_resource_group(resource_group = configuration.resource_group)
+        Azure::Armrest::ResourceGroupService.new(configuration).get(resource_group)
       end
 
+      alias resource_group_info get_resource_group
```
bzwei: completely remove `get_resource_group` aka `resource_group_info`?

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,254 @@
...
+      #
+      # Used to identify unique sessions.
+      #
+      def cache_key
+        [tenant_id, client_id, client_key].join('_')
+      end
+
+      # A list of tokens used with the current set of credentials.
+      #
+      def tokens
```
bzwei: delete this method

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,254 @@
...
+        @providers = fetch_providers
+        set_provider_info(@providers) if self.class.providers.empty?
+      end
+
+      # A combination of grant_type, tenant_id, client_id and client_key,
+      # as a single string joined by underscores.
+      #
+      # Used to identify unique sessions.
+      #
+      def cache_key
```
bzwei: make it private

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,254 @@
...
+      # Used to store unique token information.
+      @tokens = Hash.new { |hash, key| hash[key] = [] }
+
+      # A lookup table for getting api-version strings per provider and service.
+      @providers = Hash.new { |hash, key| hash[key] = {} }
+
+      # Singleton methods for the class instance variables
+      class << self
+        attr_reader :tokens
+        attr_reader :providers
```
bzwei: rename `providers` (and `@providers`) to something distinguishable from the instance method.
Both `tokens` and `providers` should be private. You can use `Configuration.send` to access them from the service classes.

lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -17,14 +17,17 @@ def initialize(configuration, options = {})
...
 
       # Return a list of available VM series (aka sizes, flavors, etc), such
-      # as "Basic_A1", though information is included as well.
+      # as "Basic_A1", though other information is included as well.
       #
       def series(location)
-        unless @@providers_hash[provider.downcase] && @@providers_hash[provider.downcase]['locations/vmsizes']
-          raise ArgumentError, "Invalid provider '#{provider}'"
+        namespace = 'microsoft.compute'
+        resources = Azure::Armrest::Configuration.providers[namespace]
```
bzwei: As mentioned above you can send to get private `providers`
Service class already have method `resources`. Are they the same? If not, better rename `resources` variable here.

lib/azure/armrest/armrest_service.rb
bzwei: Understood the  backwards compatibility reason. Can we add deprecated warning as in the example?

djberg96:
@bzwei I think everything is ready. Please let me know if there are any other changes you want to see.

bronaghs:
@djberg96 - will the MIQ code have to be updated to be compatible?

djberg96:
@bronaghs The one breaking change is that subscription ID's are mandatory, so yes. You may also see some deprecation warnings.

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,256 @@
...
+          rp.resource_types.each do |rt|
+            if rt.api_versions.any? { |v| v !~ /preview/i && Time.parse(v).utc <= Time.now.utc }
+              api_version = rt.api_versions.reject do |version|
+                version =~ /preview/i || Time.parse(version).utc > Time.now.utc
+              end.first
+            else
+              api_version = rt.api_versions.first
+            end
+
+            self.class.provider_cache[rp.namespace.downcase].merge!(rt.resource_type.downcase => api_version)
```
bzwei: We used to have `namespace` naming conflict, can you double check it is no longer a problem running at miq env.

Since this cache is used solely for api_version look up, you may want to rename `provider_cache` to something like `api_version_cache` or `api_version_lookup`

lib/azure/armrest/configuration.rb
bzwei: Need to be private

lib/azure/armrest/configuration.rb
```diff
@@ -0,0 +1,256 @@
...
+        unless client_id && client_key && tenant_id && subscription_id
+          raise ArgumentError, "client_id, client_key, tenant_id and subscription_id must all be specified"
+        end
+
+        # Allows for URI objects or Strings.
+        @proxy = @proxy.to_s if @proxy
+
+        # A one-time lookup table set at the class level
+        @providers = fetch_providers
+        set_provider_info(@providers) if self.class.provider_cache.empty?
```
bzwei: Seems a problem since `@providers` is per instance but `provider_cache` is per class while it is populated only once. Looks like `@providers` should be per class also.

lib/azure/armrest/armrest_service.rb
djberg96: This is done.

lib/azure/armrest/configuration.rb
djberg96: cache_key is now private.

lib/azure/armrest/configuration.rb
djberg96: I thought we fixed the name conflict issue, but I will double check.

How about "provider_version_cache" since the key is the provider name?

lib/azure/armrest/configuration.rb
djberg96: The @providers stores model objects, while the cache is just a simple hash for api-version lookup. I think it's handy to have both, plus the singleton method could clear the class level cache, but we would still have the instance level information.

lib/azure/armrest/configuration.rb
bzwei: `provider_api_version_cache`? This is not really provider version but api version

PR #161, opened 2016-04-15
Add the custom attribute "subscription_id" to the base model

djberg96:
This PR adds the BaseModel#subscription_id method.

This will be useful if you're iterating over resources collected for multiple subscriptions and need to distinguish between subscriptions.

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -69,6 +69,12 @@ module Azure
 
       attr_writer :resource_group
 
+      def subscription_id
+        @subscription_id ||= id[/subscriptions\/(.+?)\//i, 1] rescue nil
+      end
+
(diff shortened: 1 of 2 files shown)


djberg96:
Hm, I think this is causing an issue for configuration or subscription objects.

djberg96:
Yep, there was an issue for Subscription objects that already had a subscription_id method, so I defer to _subscription_id if it already exists.

djberg96:

djberg96:
@bronaghs @bzwei What do you think?

bzwei:
@djberg96 will this really useful if the newer release always requires a subscription id?

djberg96:
@bzwei Hm, good point. I guess this would have been useful for 0.2.x, but since we're trying to stick with 0.3.x going forward, I think I'll withdraw the PR.

PR #162, opened 2016-04-22
Use latest api-version string for StorageAccountService

djberg96:
In the past we hard coded the api-version string for the StorageAccountService class because certain versions did not work. However, the latest version (2016-01-01) works fine and even provides additional information, so I've removed the '2015-05-01-preview' version setting.

However, this did cause some changes. First and foremost the ```list_account_keys``` and ```regenerate_storage_account_keys``` methods had to be updated since they no longer return a simple hash. Instead, they now return a list of StorageAccountKey objects.

To facilitate backwards compatibility, I've added 3 simple methods to the StorageAccountKey model - the key, key1 and key2 methods. Now, instead of calling ```hash#fetch('key1')```, it's just ```obj.key1``` or even just ```obj.key```, with the latter returning key1 if present, or key2 if key1 is not present.

Luckily I think these are rarely used methods in practice and so this should have little impact out in the world.

Internally, the only other method I found that needed modification was the ```list_private_images``` method, which was a minor one-line modification.

For get/list/create/update, there were some changes in the returned json output. Mostly the fields are the same. However, with the new api-version the additional fields "encryption", "sku", and "kind" are available, while the "properties.accountType" was removed. Instead it is now under "sku.name". So, any instances of ```acct.properties.account_type``` would need to be changed to ```acct.sku.name```. I did not see any instances of that in our cloud provider code, but others might be using it. Worst case, they could always explicitly set the api_version value.

I've also included some basic specs for the StorageAccountKey class.

See https://msdn.microsoft.com/en-us/library/azure/mt163553.aspx for more details.

### DIFF
--- a/lib/azure/armrest/storage_account_service.rb
+++ b/lib/azure/armrest/storage_account_service.rb
@@ -16,7 +16,6 @@ module Azure
       # Creates and returns a new StorageAccountService (SAS) instance.
       #
       def initialize(configuration, options = {})
-        options = {'api_version' => '2015-05-01-preview'}.merge(options) # Must hard code for now
         super(configuration, 'storageAccounts', 'Microsoft.Storage', options)
       end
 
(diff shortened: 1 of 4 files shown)


djberg96:

@jerryk55 @roliveri Please take a look to see if you will be impacted by this change.

roliveri:
@djberg96 - Yes, the change to `list_account_keys` will break our code.
To facilitate backward compatibility, it may be safer to re-implement  `list_account_keys` to
return simple hashes as before, and implement a new method that provides the new behavior.
Then, we wouldn't have to worry about coordinating the release of the next version of
`azure-armrest` with the requisite changes to manageiq.

djberg96:
@roliveri Ok, I implemented separate versions for both the list_account_keys and regenerate_account_keys methods.

I went one step further and added handling for older api-version strings as well for the current methods.

lib/azure/armrest/storage_account_service.rb
```diff
@@ -108,33 +107,81 @@ def create(account_name, rgroup = configuration.resource_group, options)
...
+        parse_account_keys_from_hash(hash)
+      end
+
+      # Returns a list of StorageAccountKey objects consisting of information
+      # the primary and secondary keys. This method requires an api-version
+      # string of 2016-01-01 or later, or an error is raised.
+      #
+      # If you want a plain hash, use the list_account_keys method instead.
+      #
+      def list_account_keys2(account_name, group = configuration.resource_group)
```
roliveri: @djberg96 This is just a nit, but `list_account_key_objects` might be a better name for this method.

djberg96:
@roliveri Ok, renamed the methods.

lib/azure/armrest/storage_account_service.rb
```diff
@@ -108,33 +107,85 @@ def create(account_name, rgroup = configuration.resource_group, options)
...
 
-      # Returns the primary and secondary access keys for the given
-      # storage account as a hash.
+      # Returns the primary and secondary access keys for the given storage
+      # account. This method will return a hash with 'key1' and 'key2' as its
+      # keys.
+      #
+      # If you want a list of StorageAccountKey objects, then use the
+      # list_account_keys2 method instead.
       #
```
bronaghs: @djberg96 - looks like this method name needs updated too

lib/azure/armrest/storage_account_service.rb
```diff
@@ -108,33 +107,85 @@ def create(account_name, rgroup = configuration.resource_group, options)
...
+        validate_resource_group(group)
+
+        options = {'keyName' => key_name}
+
+        url = build_url(group, account_name, 'regenerateKey')
+        response = rest_post(url, options.to_json)
+        hash = JSON.parse(response.body)
+
+        parse_account_keys_from_hash(hash)
       end
```
bronaghs: @djberg96 - where are you calling this method?

djberg96: @bronaghs The regenerate_account_keys method? I'm not, it's a public facing method that users can call.

lib/azure/armrest/storage_account_service.rb
djberg96: @bronaghs Fixed, thanks.

roliveri:
LGTM

bronaghs:
@bzwei - Any comments before I merge?

bzwei:
@bronaghs I'll review this morning.

lib/azure/armrest/storage_account_service.rb
```diff
@@ -146,7 +197,11 @@ def list_private_images(group = configuration.resource_group)
 
         list(group).each do |lstorage_account|
           threads << Thread.new(lstorage_account) do |storage_account|
-            key = list_account_keys(storage_account.name, group).fetch('key1')
+            if Time.parse(api_version).utc >= Time.parse('2016-01-01').utc
```
bzwei: `Time.parse(api_version).utc >= Time.parse('2016-01-01').utc` appears multiple times in this class. Please extract it to a dedicated method with meaningful name.

lib/azure/armrest/storage_account_service.rb
```diff
@@ -146,7 +197,11 @@ def list_private_images(group = configuration.resource_group)
...
+              key = list_account_key_objects(storage_account.name, group).first.key
+            else
+              key = list_account_keys(storage_account.name, group).fetch('key1')
```
bzwei: I trusted you have tested this. Just curious, `key` is either a `Hash` or a `Model`, can it properly work when it later is used by `storage_account`?

djberg96: Yes, did some local testing, but feel free to double check.

The .key method just returns a string. So, `.fetch('key1')` and `.key` are basically the same. The only difference is that the `.key` method will use key1 if present, and fall back to key2 if not present, so it's slightly more robust.

lib/azure/armrest/storage_account_service.rb
djberg96: Agreed.

lib/azure/armrest/storage_account_service.rb
```diff
@@ -212,6 +267,24 @@ def parse_uri(uri)
 
       private
 
+      # Check to see if the api-version string is 2016-01-01 or later.
+      def recent_api_version
```
bzwei: `recent_api_version?`

djberg96: Sure. I'll push that up.

djberg96:
@bzwei @bronaghs Note that I am ignoring these last few rubocop warnings.

PR #163, opened 2016-04-29
Remove hard coded api-version string for TemplateDeploymentService

djberg96:
This removes the hard-coded api-version string for TemplateDeploymentService. Our changes in our api-version handler make this unnecessary. The current version is 2016-02-01.

I also updated the specs to use the current api-version string.

As far as I can tell, there have been no changes to the output. I tested all the get and list methods with some basic scripting and they seemed to work fine.


djberg96:
@bzwei Please review.

bzwei:
Looks good. Please fix the align hash style then it can be merged.

djberg96:
@bzwei I already fixed it. The bot mailer was really slow for some reason.

PR #164, opened 2016-05-06
Modify the fetch_subscription_id method so that it skips over unauthorized subscriptions

djberg96:
Users will sometimes see "The access token is from the wrong issuer". This can happen if there are multiple AD's (tenants) on the same subscription, but then try to access data from a subscription that belongs to a separate tenant.

In our case the tenant used in the fetch_token method would be mismatched with a subscription picked up in the fetch_subscription_id method.

So, this modifies the fetch_subscription_id method to make a test request (looking for something with a relatively small response - tags) for each subscription. If it's invalid, it skips it and looks for the next one. It breaks out of the loop the first time it finds a valid one that is also enabled.

This will, of course, be a moot point in 0.3.0, but we should fix this for the 0.2.x branch.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -171,25 +171,49 @@ module Azure
 
         url = File.join(Azure::Armrest::RESOURCE, "subscriptions?api-version=#{config.api_version}")
 
-        response = rest_get(
+        options = {
           :url         => url,
           :proxy       => config.proxy,
           :ssl_version => config.ssl_version,
           :ssl_verify  => config.ssl_verify,
-          :headers => {
+          :headers     => {
             :content_type  => config.content_type,
             :authorization => config.token
           }
-        )
+        }
 
-        array = JSON.parse(response)["value"]
+        response = rest_get(options)
+        array = JSON.parse(response)['value']
 
         if array.nil? || array.empty?
           raise ArgumentError, "No associated subscription found"
         end
 
+        results = []
+
+        # Skip over any invalid subscriptions
+        array.each do |sub_hash|
+          sub_id = sub_hash['subscriptionId']
+
+          # Use tagNames as a test URL
+          test_url = File.join(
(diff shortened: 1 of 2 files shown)


djberg96:

I realize that this will be a moot point in 0.3.x, but we should fix it for 0.2.x.

lib/azure/armrest/armrest_service.rb
```diff
@@ -171,25 +171,37 @@ def self.fetch_subscription_id(config)
...
+            next
+          else
+            results << sub_hash
+            break if sub_hash['state'] == 'Enabled'
+          end
+        end
+
         # Look for the first enabled subscription, otherwise just take the first subscription found.
-        hash = array.find{ |h| h['state'] == 'Enabled' } || array.first
+        hash = results.find { |h| h['state'] == 'Enabled' } || results.first
```
bronaghs: @djberg96  - Do we need this line? 
Will the results array only contain 1 subscription given the `break` statement?

djberg96: It's possible that there could be disabled subscriptions in the array as well.

lib/azure/armrest/armrest_service.rb
```diff
@@ -412,8 +424,24 @@ def self.raise_api_exception(e)
...
+          :ssl_verify  => config.ssl_verify,
+          :headers => {
+            :content_type  => config.content_type,
+            :authorization => config.token
+          }
+        )
+      end
+
+      private_class_method :get_response
+
```
bronaghs: @djberg96  Bit of a duplication of `rest_execute`

djberg96: Eh, I'm wrapping the options because they're identical between calls.

lib/azure/armrest/armrest_service.rb
bronaghs: OK, makes sense.

PR #166, opened 2016-05-11
Fix private rest method calls in StorageAccount model.

djberg96:
Forgot some internal methods used in the StorageAccount model that needed to be updated for 0.3.x.


djberg96:
@bzwei Please review and merge.

PR #167, opened 2016-05-11
Hide cache methods in Configuration class

bzwei:
Previously the whole caches are exposed. Now only provide query methods
to access cached data (for api-version and token)

### DIFF
--- a/lib/azure/armrest/configuration.rb
+++ b/lib/azure/armrest/configuration.rb
@@ -1,25 +1,25 @@
 module Azure
   module Armrest
     class Configuration
-      # Used to store unique token information.
-      @token_cache = Hash.new { |hash, key| hash[key] = [] }
-
-      # A lookup table for getting api-version strings per provider and service.
-      @provider_version_cache = Hash.new { |hash, key| hash[key] = {} }
-
-      # Singleton methods for the class instance variables
-      class << self
-        attr_reader :token_cache
-        attr_reader :provider_version_cache
-
-        # Clear all class level caches. Typically used for testing only.
-        def clear_caches
-          @token_cache    = Hash.new { |h, k| h[k] = [] } # Token caches
-          @provider_version_cache = Hash.new { |h, k| h[k] = {} } # Providers
-        end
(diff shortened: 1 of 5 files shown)


lib/azure/armrest/configuration.rb
```diff
@@ -193,7 +189,10 @@ def validate_token_time(time)
       # a non-preview version that is not set in the future. Otherwise, just
       # just the most recent one.
       #
-      def set_provider_info(providers)
+      def set_provider_api_versions
+        # A lookup table for getting api-version strings per provider and service.
+        @provider_api_versions = Hash.new { |hash, key| hash[key] = {} }
```
djberg96: Shouldn't the @provider_api_versions initialization happen in the constructor?

lib/azure/armrest/configuration.rb
```diff
@@ -86,39 +83,51 @@ def clear_caches
...
-        # A one-time lookup table set at the class level
         @providers = fetch_providers
-        set_provider_info(@providers) if self.class.provider_version_cache.empty?
+        set_provider_api_versions
+      end
+
+      def hash
+        [tenant_id, client_id, client_key].join('_').hash
+      end
+
```
djberg96: Should the hash method be private?

lib/azure/armrest/configuration.rb
bzwei: No need. `set_provider_info` is called by the constructor anyway. Technically if this method is called separately (which does not happen), `set_provider_info` needs to be reset here.

lib/azure/armrest/configuration.rb
bzwei: `hash` and `eql?` are twin overridden methods from `Object`, similar to `hashCode` and `equals` in Java.

lib/azure/armrest/configuration.rb
djberg96: I'm just wondering if that will cause a warning in verbose mode. We can always revisit.

djberg96:
Looks good. I guess it would be best to fix those rubocop TZ warnings first.

PR #171, opened 2016-05-18
Rename clear_caches to clear_token_cache, and update specs.

djberg96:
Since we only have one class level cache now, the `clear_caches` method is a bit of a misnomer. This just renames it to `clear_token_cache`. It also creates an alias for backwards compatibility, as well as a deprecation notice.

I used the `class << self` syntax, since it's easier to alias singleton methods that way, and gets rid of `self` everywhere, but can undo that if that style is not preferred.

### DIFF
--- a/lib/azure/armrest/configuration.rb
+++ b/lib/azure/armrest/configuration.rb
@@ -1,25 +1,32 @@
 module Azure
   module Armrest
     class Configuration
-      # Clear all class level caches. Typically used for testing only.
-      def self.clear_caches
-        # Used to store unique token information.
-        @token_cache = Hash.new { |h, k| h[k] = [] }
+      class << self
+        extend Gem::Deprecate
+
+        # Clear all class level caches. Typically used for testing only.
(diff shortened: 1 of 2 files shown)


djberg96:
@bzwei Please review.

bzwei:
I left it as `clear_caches` on purpose. Yes right now we only have one cache. But if in the future we need to add more to caches the client call to `clear_caches` does not need to change.

I am ok with the `class << self` style.

djberg96:
@bzwei I could remove the deprecation notice, and just leave the alias. If we ever do add more class level cache variables, it would be a simple matter of redefining clear_token_cache.

bzwei:
`clear_caches` is the existing method that is already in use. If we introduce `clear_token_cache`. Now we have two methods. Chances are both methods are in use even though they are the same. When the time comes that `clear_token_cache` is no longer meaningful, we cannot retire it because it is in use.

More importantly caching is internal. We don't need to tell the user token is cached.

Issue #172, opened 2016-05-20
Delegated method list_providers not found


djberg96:
The delegated :list_providers method in Azure::Armrest::ArmrestService (on master) doesn't appear to work:
```
conf = Azure::Armrest::Configuration.new(creds)
vms  = Azure::Armrest::VirtualMachineService.new(conf)

p vms.list_providers
```
Result:
```
/Users/dberger/Repositories/azure-armrest-djberg96/lib/azure/armrest/armrest_service.rb:50:in `list_providers': undefined method `list_providers' for #<Azure::Armrest::Configuration:0x007fad6cd3dbe8> (NoMethodError)
	from locations.rb:11:in `<main>'
```
I also tried the alias :providers, but that also failed:
```
>ruby locations.rb 
NOTE: Azure::Armrest::VirtualMachineService#providers is deprecated; use list_providers instead. It will be removed on or after 2018-01-01.
Azure::Armrest::VirtualMachineService#providers called from locations.rb:11.
/Users/dberger/Repositories/azure-armrest-djberg96/lib/azure/armrest/armrest_service.rb:50:in `list_providers': undefined method `list_providers' for #<Azure::Armrest::Configuration:0x007f8bbb758200> (NoMethodError)
	from /Users/dberger/.rbenv/versions/2.2.3/lib/ruby/site_ruby/2.2.0/rubygems/deprecate.rb:63:in `block (2 levels) in deprecate'
	from locations.rb:11:in `<main>'
```

PR #174, opened 2016-05-24
Fix delegation bug

djberg96:
This addresses https://github.com/ManageIQ/azure-armrest/issues/172

Basically, we can't delegate "list_providers" because that method doesn't exist in the Configuration class, and it wouldn't make sense to rename it, since it's just an instance variable.

I know we wanted to be consistent with the naming, but I think we should let this one slide. Also, it's how the current API works, so there's no breakage.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -47,7 +47,7 @@ module Azure
       # Returns a list of the available resource providers. This is really
       # just a wrapper for Azure::Armrest::Configuration#providers.
       #
-      delegate :list_providers, :to => :configuration
+      delegate :providers, :to => :configuration, :prefix => :list
 
       alias providers list_providers
       deprecate :providers, :list_providers, 2018, 1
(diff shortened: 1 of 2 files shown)


djberg96:
@bzwei Please review.

lib/azure/armrest/armrest_service.rb
```diff
@@ -47,10 +47,7 @@ def initialize(armrest_configuration, service_name, default_provider, options)
       # Returns a list of the available resource providers. This is really
       # just a wrapper for Azure::Armrest::Configuration#providers.
       #
-      delegate :list_providers, :to => :configuration
-
-      alias providers list_providers
-      deprecate :providers, :list_providers, 2018, 1
+      delegate :providers, :to => :configuration
```
bzwei: You can try
`delegate :providers, :to => :configuration, :prefix => :list`

djberg96: Oh, I didn't know you could do that. I'll update the PR.

PR #176, opened 2016-06-01
Add UsageService class, specs, and related changes.

djberg96:
This PR adds a UsageService class that people (including us) can potentially use to track resource consumption.

https://msdn.microsoft.com/en-us/library/azure/mt219001

### DIFF
--- a/lib/azure/armrest/billing/usage_service.rb
+++ b/lib/azure/armrest/billing/usage_service.rb
@@ -1,0 +1,54 @@
+module Azure
+  module Armrest
+    module Billing
+      class UsageService < ArmrestService
+        # Creates and returns a new UsageService object.
+        #
+        def initialize(configuration, options = {})
+          options = options.merge('api_version' => '2015-06-01-preview')
+          super(configuration, 'subscriptions', 'Microsoft.Commerce', options)
+        end
+
+        # List usage details. The +options+ hash may include the following
+        # filters:
+        #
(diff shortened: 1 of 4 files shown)


djberg96:
@bzwei Please review.

lib/azure/armrest/billing/usage_service.rb
```diff
@@ -0,0 +1,56 @@
+require 'cgi'
```
bzwei: where is this library used?

lib/azure/armrest/billing/usage_service.rb
```diff
@@ -0,0 +1,56 @@
...
+
+module Azure
+  module Armrest
+    module Billing
+      class UsageService < ArmrestService
+        # Creates and returns a new UsageService object.
+        #
+        def initialize(configuration, options = {})
+          options['api_version'] = '2015-06-01-preview' # Must hard code for now
```
bzwei: you may want to do a `merge`, leaving the argument `options` unmodified

lib/azure/armrest/billing/usage_service.rb
djberg96: Ah, forgot to remove it. Was thinking I had to encode the dates, but rest-client already does it.

djberg96:
@bzwei I've cleaned it up.

Issue #178, opened 2016-06-01
Library should store response headers somehow


djberg96:
Currently we basically ignore the response headers. However, they may contain useful information:

```
{:cache_control=>"no-cache", :pragma=>"no-cache", :content_type=>"application/json", :content_encoding=>"gzip", :expires=>"-1", :vary=>"Accept-Encoding", :x_ms_request_id=>"a7e0e1ad-6f3a-4eeb-ace8-9c77c3377179", :server=>"Microsoft-Azure-Storage-Resource-Provider/1.0, Microsoft-HTTPAPI/2.0", :x_ms_ratelimit_remaining_subscription_reads=>"14872", :x_ms_correlation_request_id=>"a7e0e1ad-6f3a-4eeb-ace8-9c77c3377179", :x_ms_routing_request_id=>"NORTHCENTRALUS:20160601T182659Z:a7e0e1ad-6f3a-4eeb-ace8-9c77c3377179", :strict_transport_security=>"max-age=31536000; includeSubDomains", :date=>"Wed, 01 Jun 2016 18:26:59 GMT", :connection=>"close"}
```

Notably, the `:x_ms_ratelimit_remaining_subscription_reads` would be useful to know, especially if we get close to zero.

I'm not sure how these should be stored, though. Storing header information in the model objects would be wasteful since it would be identical in every object. Perhaps a singleton that stores the header information? I'm open to suggestions.

bzwei:
How are we going to use this information? If we are not sure at this moment, we should postpone the design. I am wondering the header info could be per subscription, per service type, or per call, so it is hard to decide where to store. In addition the header contains multiple keys, each could have different scope, making it harder to be processed in a single way.
I think we can parse the header as needed, not blindly store them all.

djberg96:
@bzwei I think the plan is to use as a way to throttle requests, or rescue failed requests, when the limit is approaching zero.

Perhaps we could leverage the ArmrestCollection class that we added originally for events and continuation token storage, and add a `:headers` attribute. While this would only work for `list` and `list_all` methods, and think that's where it would be most useful anyway, and it would mean storing it once per collection set, rather than once per model.

djberg96:
While I think this is still a good idea, the main culprit is list_private_images. What we need is list_all_private_images, which would reduce multiple calls (one per resource group) down to a single call.

djberg96:
https://github.com/ManageIQ/azure-armrest/pull/180

Issue #154, opened 2016-03-28
Deal with $skip tokens in a uniform manner


djberg96:
The recent events stuff has forced us to reconsider how we're wrapping some of the results for various list methods. The problem, in general, is that if you specify a filter of any sort that limits the results, the user may want to get more results from that point on, but there's no way to do that at the moment. This is because the 'nextLink' attribute is not included in the results that we parse.

The solution for Insight::Events that we use, for the moment anyway, is to return an EventList object. It contains the list results (model objects) as one attribute, and the skip token as another.

A couple of ideas to deal with this are:

1) Always return an XList object for all list and list_all methods.

Advantages: It's a clean design. This would make things consistent across the board.
Disadvantages: Causes major breakage. Lots of code would have to be updated for people already using the gem.

2) Return an XList object only if a skip token is found. Otherwise, return a regular list of model objects as we do now.

Advantages: Wouldn't cause breakage for most methods since the 200 limit that applies to events doesn't seem to apply to other resources. Only a filter supplied by the users would generate a skip token.

Disadvantages: Returning different objects based on context is gross. And we couldn't guarantee there wouldn't be any surprises, because it's possible other resources have a default limit that we don't know about.

3) Just add the skip token property on ALL model objects instead of returning a new type of object.

Advantages: This requires minimal change to the code base, and causes no breakage.

Disadvantages: Lots of duplication in the model objects, since it would be identical for all model objects in a particular result set.

Personally I'm leaning towards the last option at the moment, but am open to suggestions.

bzwei:
3) :+1: 
May consider only add skip token to the last object in the list if duplication is a concern.

djberg96:
@bzwei True, we could get the object list first and then do `objects.last['nextLink'] = next_link` within the list and list_all methods.

durandom:
This is how aws does it (see `next_marker`) 
http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Client.html#list_objects-instance_method

And here for the API:
http://docs.aws.amazon.com/sdkforruby/api/index.html#Paging_Responses

So, they dont have a `list` and `list_all` method, but  handle the paging inside the sdk in case you are using an enumerator or via `.next_page` if not.

I would go for 3) too. But include it in every object returned. In this scenario the user has to handle the paging anyhow, so he has to check if he hit the last object and trigger another `list` request with the skip marker. I guess this does not break the current API and can be released as a minor version.

[...]

PR #180, opened 2016-06-02
Modify the ResourceGroupBasedService to return ArmrestCollection objects.

djberg96:
This is a solution to https://github.com/ManageIQ/azure-armrest/issues/178.

This modifies the ResourceGroupBasedService so that the `list` and `list_all` methods return an `ArmrestCollection` object instead of a plain array. In turn, the `ArmrestCollection` object has been modified to include a `response_headers` accessor, and a `ResponseHeaders` model has been added as well, which is then set within the `list` and `list_all` methods.

*Update: the ResponseHeaders class was already added within a separate PR, so we use that now. Also, the method was changed from :headers to :response_headers.*

The short version is that now you can do something like this:

```
results = vms.list_all
results.response_headers.x_ms_ratelimit_remaining_subscription_reads
```

This lets us store the returned header information per collection set, rather than per object, since such information would be redundant.

*Update: I should amend that to say that we want the header information set once per request. So, a :list or :list_all call would set it once for the collection, while a :get would set it once on the returned object.*

~~A couple specs are currently failing because of missing attributes, but they're minor.~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /pkg/
 /spec/reports/
 /tmp/
+/futz/
--- a/lib/azure/armrest/armrest_collection.rb
+++ b/lib/azure/armrest/armrest_collection.rb
@@ -4,6 +4,38 @@ module Azure
   module Armrest
     class ArmrestCollection < Array
       attr_accessor :continuation_token
+      attr_accessor :response_headers
+
+      # Creates and returns a ArmrestCollection object based on JSON input,
+      # using +klass+ to generate the list elements. In addition, both the
+      # response headers and continuation token are set.
+      #
+      # You may optionally pass a plain array instead, in which case you
+      # get back an array re-wrapped as an ArmrestCollection object, but
+      # the response_headers and continuation_token properties are not set
+      # automatically.
+      #
+      def initialize(response, klass = nil)
+        if response.kind_of?(Array)
+          super(response)
+        else
+          json_response = JSON.parse(response)
+          @response_headers = response.headers
+          @continuation_token = parse_skip_token(json_response)
+          super(json_response['value'].map { |hash| klass.new(hash) })
+        end
+      end
+
+      alias skip_token continuation_token
+      alias skip_token= continuation_token=
+
+      private
+
+      # Parse the skip token value out of the nextLink attribute from a response.
+      def parse_skip_token(json)
+        return nil unless json['nextLink']
+        json['nextLink'][/.*?skipToken=(.*?)$/i, 1]
+      end
     end
   end
 end
--- a/lib/azure/armrest/armrest_collection.rb
+++ b/lib/azure/armrest/armrest_collection.rb
@@ -6,35 +6,31 @@ module Azure
       attr_accessor :continuation_token
       attr_accessor :response_headers
 
-      # Creates and returns a ArmrestCollection object based on JSON input,
-      # using +klass+ to generate the list elements. In addition, both the
-      # response headers and continuation token are set.
-      #
-      # You may optionally pass a plain array instead, in which case you
-      # get back an array re-wrapped as an ArmrestCollection object, but
-      # the response_headers and continuation_token properties are not set
-      # automatically.
-      #
-      def initialize(response, klass = nil)
-        if response.kind_of?(Array)
-          super(response)
-        else
-          json_response = JSON.parse(response)
-          @response_headers = response.headers
-          @continuation_token = parse_skip_token(json_response)
-          super(json_response['value'].map { |hash| klass.new(hash) })
-        end
-      end
-
       alias skip_token continuation_token
       alias skip_token= continuation_token=
 
-      private
+      class << self
+        # Creates and returns a ArmrestCollection object based on JSON input,
+        # using +klass+ to generate the list elements. In addition, both the
+        # response headers and continuation token are set.
+        #
+        def create_from_response(response, klass = nil)
+          json_response = JSON.parse(response)
+          array = new(json_response['value'].map { |hash| klass.new(hash) })
 
-      # Parse the skip token value out of the nextLink attribute from a response.
-      def parse_skip_token(json)
-        return nil unless json['nextLink']
-        json['nextLink'][/.*?skipToken=(.*?)$/i, 1]
+          array.response_headers = response.headers
+          array.continuation_token = parse_skip_token(json_response)
+
+          array
+        end
+
+        private
+
+        # Parse the skip token value out of the nextLink attribute from a response.
+        def parse_skip_token(json)
+          return nil unless json['nextLink']
+          json['nextLink'][/.*?skipToken=(.*?)$/i, 1]
+        end
       end
     end
   end
--- a/lib/azure/armrest/insights/event_service.rb
+++ b/lib/azure/armrest/insights/event_service.rb
@@ -53,7 +53,7 @@ module Azure
           response = rest_get(url)
 
           klass  = Azure::Armrest::Insights::Event
-          events = Azure::Armrest::ArmrestCollection.new(response, klass)
+          events = Azure::Armrest::ArmrestCollection.create_from_response(response, klass)
 
           if options[:all] && events.continuation_token
             events.push(*list(options.merge(:skip_token => events.continuation_token)))
--- a/lib/azure/armrest/insights/event_service.rb
+++ b/lib/azure/armrest/insights/event_service.rb
@@ -44,22 +44,16 @@ module Azure
         #   filter = "eventTimestamp ge #{date} and eventChannels eq 'Admin, Operation'"
         #   select = "resourceGroupName, operationName"
         #
-        #   ies.list(:filter => filter, :select => select, :all => true).events.each{ |event|
+        #   ies.list(:filter => filter, :select => select, :all => true).each{ |event|
         #     p event
         #   }
         #
         def list(options = {})
           url = build_url(options)
           response = rest_get(url)
-          json_response = JSON.parse(response.body)
 
-          events = ArmrestCollection.new(
-            json_response['value'].map do |hash|
-              Azure::Armrest::Insights::Event.new(hash)
-            end
-          )
-
-          events.continuation_token = parse_skip_token(json_response)
+          klass  = Azure::Armrest::Insights::Event
+          events = Azure::Armrest::ArmrestCollection.new(response, klass)
 
           if options[:all] && events.continuation_token
             events.push(*list(options.merge(:skip_token => events.continuation_token)))
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -22,6 +22,8 @@ module Azure
 
       attr_hash :tags
 
+      attr_accessor :response_headers
+
       # Constructs and returns a new JSON wrapper class. Pass in a plain
       # JSON string and it will automatically give you accessor methods
       # that make it behave like a typical Ruby object. You may also pass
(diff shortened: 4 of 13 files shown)


djberg96:
@bzwei What do you think?

lib/azure/armrest/armrest_collection.rb
```diff
@@ -4,6 +4,7 @@ module Azure
   module Armrest
     class ArmrestCollection < Array
       attr_accessor :continuation_token
+      attr_accessor :response_headers
```
bzwei: Add an initializer to accept the response_headers as an optional hash, and covert to model object internally. Then the calling can be greatly simplified.

lib/azure/armrest/model/base_model.rb
```diff
@@ -22,6 +22,8 @@ def self.attr_hash(*attrs)
 
       attr_hash :tags
 
+      attr_accessor :response_headers
```
bzwei: Similarly add response_headers as an optional hash argument in the initializer.
How useful are the headers? The model object if returned by a single GET has headers, but if in a list does not have headers. Will the difference cause trouble for application?

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -2,25 +2,56 @@ module Azure
...
+      # Returns an ArmrestCollection, with the response headers set
+      # for the operation as a whole.
+      #
       def list(rgroup = configuration.resource_group)
         validate_resource_group(rgroup)
 
         url = build_url(rgroup)
         url = yield(url) || url if block_given?
+
         response = rest_get(url)
```
bzwei: I would expect all LIST methods from Azure API to have `continuation_token`. Otherwise how do they manage a super long list without paging?

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -35,19 +66,34 @@ def list(rgroup = configuration.resource_group)
...
-        results = JSON.parse(response)['value'].map { |hash| model_class.new(hash) }
+        json_response = JSON.parse(response)['value']
+
+        results = ArmrestCollection.new(json_response.map { |hash| model_class.new(hash) })
+
         filter.empty? ? results : results.select { |obj| filter.all? { |k, v| obj.public_send(k) == v } }
+
+        results.response_headers = Azure::Armrest::ResponseHeaders.new(response.headers)
```
bzwei: You should use new `results` from above line. But it is no longer needed if you refactor the initializer as suggested.

lib/azure/armrest/model/base_model.rb
djberg96: We don't want to add headers for every object, because they would all be identical, and would create unnecessary memory bloat. It's meant to only be set once per request.

lib/azure/armrest/armrest_collection.rb
djberg96: I generally avoid redefining initialize on a subclass of a core class because it can lead to unexpected behavior. Since this is only used internally, I don't see the advantage of redefining initialize.

lib/azure/armrest/resource_group_based_service.rb
djberg96: I'm actually not entirely sure that Azure sets this in a uniform manner, or for all resources. Consider the way tokens are parsed for insights versus storage.

Also, I still have this issue outstanding: https://github.com/ManageIQ/azure-armrest/issues/154

Anyway, I agree in principal, but think it should be a separate PR for now.

lib/azure/armrest/armrest_collection.rb
djberg96: Upon further review, maybe this would be easier, since we have a bunch of custom list methods scattered throughout.

lib/azure/armrest/resource_group_based_service.rb
djberg96: After scouring the docs a bit more, I think we can just use parse_skip_token from armrest_service.rb for most cases.

lib/azure/armrest/model/base_model.rb
bzwei: Memory will not be a concern if every model object holds a reference to the same header instance. 

I assume the header responding for a LIST request is different from the header responding for a GET request for a single object.  I therefore agree with you that it may be wrong to assign an unprocessed LIST header to each individual model object, unless the two headers are identical.

In addition, we should hide the response_header for internal use only because of the above discussed difference. Is there a use case that the user should know the header?

lib/azure/armrest/resource_group_based_service.rb
djberg96: Agreed.

djberg96:
@bzwei I've updated the ArmrestCollection class to accept a response object and a class. It then parses and sets the information internally, and returns an array of the class type provided.

Please let me know if this is what you had in mind. There are a few event specs that are failing now that I might need some help with.

lib/azure/armrest/insights/event_service.rb
```diff
@@ -44,22 +44,16 @@ def initialize(armrest_configuration, options = {})
...
-          json_response = JSON.parse(response.body)
 
-          events = ArmrestCollection.new(
-            json_response['value'].map do |hash|
-              Azure::Armrest::Insights::Event.new(hash)
-            end
-          )
-
-          events.continuation_token = parse_skip_token(json_response)
+          klass  = Azure::Armrest::Insights::Event
```
bzwei: My thought is for `ArmrestCollection#initialize()` 
```
  def initialize(array, header = nil)
    # parse and process header
    super(array)
  end
```
then you only needs to append like this
```
events = ArmrestCollection.new(
  json_response['value'].map { |hash| Azure::Armrest::Insights::Event.new(hash) },
  response.header
)
```

djberg96: That doesn't deal with skip tokens (which are in the body), and doesn't provide a lot of added value IMO. My approach handles skip tokens and reduces a lot of code.

lib/azure/armrest/model/base_model.rb
djberg96: The response_headers need to be public IMO so that people can inspect rate limit information.

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -35,19 +61,30 @@ def list(rgroup = configuration.resource_group)
...
         url = build_url
         url = yield(url) || url if block_given?
+
         response = rest_get(url)
-        results = JSON.parse(response)['value'].map { |hash| model_class.new(hash) }
+        results  = Azure::Armrest::ArmrestCollection.new(response, model_class)
+
         filter.empty? ? results : results.select { |obj| filter.all? { |k, v| obj.public_send(k) == v } }
+
+        results
```
bzwei: remove line 69 and 70

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -104,16 +141,22 @@ def model_class
...
           response = rest_get(build_url(rg.name))
-          results = JSON.parse(response)['value'].map { |hash| model_class.new(hash) }
+          json_response = JSON.parse(response.body)['value']
+          headers = Azure::Armrest::ResponseHeaders.new(response.headers)
+          results = json_response.map { |hash| model_class.new(hash) }
           mutex.synchronize { array << results } unless results.blank?
         end
 
-        array.flatten
+        array = ArmrestCollection.new(array.flatten)
```
bzwei: will this work with the new ArmrestCollection initializer?

spec/insights_event_service_spec.rb
```diff
@@ -6,7 +6,12 @@
 require 'spec_helper'
 
 describe "Insights::EventService" do
-  before { setup_params }
+  before do
+    setup_params
+    hash = {:x_ms_ratelimit_remaining_subscription_reads => '14999'}
+    allow_any_instance_of(String).to receive(:headers).and_return(hash)
```
bzwei: `allow_any_instance_of` needs to be avoided.

lib/azure/armrest/resource_group_based_service.rb
djberg96: Done.

lib/azure/armrest/resource_group_based_service.rb
djberg96: Nope, good catch. That was something I was tinkering with and forgot to get back to.

.gitignore
```diff
@@ -7,3 +7,4 @@
 /pkg/
 /spec/reports/
 /tmp/
+/futz/
```
bzwei: where does this come from?

lib/azure/armrest/armrest_collection.rb
```diff
@@ -4,6 +4,32 @@ module Azure
...
+        array.response_headers = response.headers
+        array.continuation_token = parse_skip_token(json_response)
+
+        array
+      end
+
+      alias skip_token continuation_token
+      alias skip_token= continuation_token=
+
+      private
```
bzwei: @djberg96 please fix. `private` does not make class method private. See the rubocop messages. Those should be addressed.

lib/azure/armrest/insights/event_service.rb
```diff
@@ -44,22 +44,16 @@ def initialize(armrest_configuration, options = {})
...
+          events = Azure::Armrest::ArmrestCollection.create_from_response(response, klass)
```
bzwei: can line 55 and 56 combine into one line?

lib/azure/armrest/model/storage_account.rb
```diff
@@ -99,11 +99,12 @@ def table_data(name, key = nil, options = {})
         key ||= properties.key1
 
         query = build_query(options)
-
         response = table_response(key, query, name)
-        json_response = JSON.parse(response.body)
 
-        data = ArmrestCollection.new(json_response['value'].map { |t| TableData.new(t) })
+        klass = Azure::Armrest::StorageAccount::TableData
+        data  = Azure::Armrest::ArmrestCollection.create_from_response(response, klass)
```
bzwei: same here, better one line?

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -2,25 +2,51 @@ module Azure
...
-        model_class.new(response) unless response.empty?
+
+        obj = nil
+
+        unless response.empty?
+          obj = model_class.new(response.body)
+          obj.response_headers = Azure::Armrest::ResponseHeaders.new(response.headers)
+        end
+
+        obj
```
bzwei: Yours is fine. Does this look better?
```
model_class.new(response).tap |obj|
  obj.response_headers = Azure::Armrest::ResponseHeaders.new(response.headers)
end unless response.empty?
```

djberg96: That's harder for me to read.

lib/azure/armrest/insights/event_service.rb
djberg96: I'm just trying to avoid extremely long lines.

lib/azure/armrest/model/storage_account.rb
djberg96: I'm just trying to avoid extremely long lines.

.gitignore
djberg96: Oops, I thought that would be ignored. Eh, it's harmless. It's just a place where I put local test scripts.

lib/azure/armrest/armrest_collection.rb
djberg96: @bzwei Fixed. It does if I use a class/self block. :)

PR #181, opened 2016-06-10
Add support for a TooManyRequests exception.

djberg96:
This just adds a TooManyRequestsException class that wraps a RestClient::TooManyRequests error. We will need this to handle errors caused by issuing too many requests in too short a time.


djberg96:
@bzwei Please review.

PR #182, opened 2016-06-14
Replace uri with addressable

djberg96:
This PR replaces the uri standard library with addressable, which I've found is simply more robust and a better library in general. This was inspired by a recent issue that the azure-signature gem hit, which I want to avoid with this gem.

I also removed all calls to URI.escape since rest-client already auto-escapes, and made one minor doc update.


djberg96:

Issue #183, opened 2016-06-16
All url sent by RestClient should be encoded


bzwei:
This is a follow-up with #182. 
All url sent by RestClient should be encoded.

djberg96:
rest-client already does I believe.

bzwei:
Unfortunately no. http://stackoverflow.com/questions/20254749/restclient-parsing-uri-with-space-on-request-path

bzwei:
fixed by #184

PR #184, opened 2016-06-16
Encode url before handing to RestClient to execute a request

bzwei:
Fixes issue #183

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -161,7 +161,11 @@ module Azure
         private
 
         def rest_execute(options, http_method = :get)
-          RestClient::Request.execute(options.merge(:method => http_method))
+          options = options.merge(
+            :method => http_method,
+            :url    => Addressable::URI.escape(options[:url])
+          )
+          RestClient::Request.execute(options)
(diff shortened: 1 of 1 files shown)


lib/azure/armrest/armrest_service.rb
```diff
@@ -161,7 +161,11 @@ class << self
         private
 
         def rest_execute(options, http_method = :get)
-          RestClient::Request.execute(options.merge(:method => http_method))
+          options = options.merge(
+            :method => http_method,
+            :url    => Addressable::URI.parse(options[:url]).normalize.to_str
```
djberg96: What does .parse + .normalize + .to_str do that .escape doesn't do?

bzwei: @djberg96 I actually don't know. I just followed what is said in  http://stackoverflow.com/questions/20254749/restclient-parsing-uri-with-space-on-request-path that "The documentation suggests to parse and normalize the URL first".
I don't know what document it referenced 😄

djberg96: Well, I'd rather not call extra methods if we don't have to. Could we add some tests?

djberg96:
👍

PR #187, opened 2016-06-20
Add StorageAccountService#list_all_private_images method

djberg96:
This PR adds the StorageAccount#list_all_private_images method.

This is similar to StorageAccount#list_private_images, except that it uses a single call to get all storage accounts, which the user can filter, prior to making additional REST calls needed to get private images.

This is much more efficient, and effectively solves a problem where users would hit the request rate limit because they were iterating over too many resource groups.

All of the shared code was moved to private methods.


djberg96:

(Somehow I picked up 27fb308, but it's harmless).

PR #188, opened 2016-06-21
Add StorageAccountService#list_all_private_images method

djberg96:
This PR adds the StorageAccount#list_all_private_images method. This supercedes #187 that got goofed up somehow.

This is similar to StorageAccount#list_private_images, except that it uses a single call to get all storage accounts, which the user can filter, prior to making additional REST calls needed to get private images.

This is much more efficient, and effectively solves a problem where users would hit the request rate limit because they were iterating over too many resource groups.

All of the shared code was moved to private methods.

### DIFF
--- a/lib/azure/armrest/storage_account_service.rb
+++ b/lib/azure/armrest/storage_account_service.rb
@@ -176,51 +176,35 @@ module Azure
 
       alias regenerate_storage_account_key_objects regenerate_account_key_objects
 
-      # Returns a list of images that are available for provisioning for all
-      # storage accounts in the provided resource group. The custom keys
-      # :uri and :operating_system have been added for convenience.
+      # Returns a list of PrivateImage objects that are available for
+      # provisioning for all storage accounts in the current subscription.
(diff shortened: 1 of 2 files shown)


djberg96:

lib/azure/armrest/storage_account_service.rb
```diff
@@ -176,51 +176,35 @@ def regenerate_account_key_objects(account_name, group = configuration.resource_
...
+      # You may optionally reduce the set of storage accounts that will
+      # be scanned by providing a filter, where the keys are StorageAccount
+      # properties.
+      #
+      # Example:
+      #
+      #   sas.list_all_private_images(:location => 'eastus', resource_group => 'some_group')
+      #
+      def list_all_private_images(filter = {})
+        storage_accounts = list_all.select { |acct| filter.all? { |k, v| acct.public_send(k) == v } }
```
bzwei: nice!

bzwei:
👍

Issue #189, opened 2016-06-21
Add a Configuration#validate method


djberg96:
At the moment, the Configuration class does not provide a way to validate credentials before they're used.

We should add a Configuration#validate (or valid?) method that makes some sort of simple REST call to ensure the credentials are actually good.

djberg96:
Currently we do at least get client_id, client_key and tenant_id validation via the fetch_providers call in the constructor. However, that does not validate the subscription.

djberg96:
Solved by https://github.com/ManageIQ/azure-armrest/pull/190.

PR #190, opened 2016-06-24
Added Configuration#validate_subscription!

djberg96:
This adds a Configuration#validate_subscription! method, and addresses https://github.com/ManageIQ/azure-armrest/issues/189.

The main issue is that the constructor effectively does a validation check for tenant ID, client ID and client key via the `fetch_providers` method, but it does not validate the subscription. This method allows users to explicitly validate the subscription and/or check to ensure that it's enabled.

### DIFF
--- a/lib/azure/armrest/configuration.rb
+++ b/lib/azure/armrest/configuration.rb
@@ -80,6 +80,10 @@ module Azure
       # Although you can specify an :api_version, it is typically overridden
       # by individual service classes.
       #
+      # Note that while the constructor will fail if invalid credentials are
+      # supplied, it does not validate your subscription ID. If you want to
+      # validate your subscription ID as well, use the validate_subscription! method.
+      #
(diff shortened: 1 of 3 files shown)


djberg96:
@bzwei First attempt. What do you think?

djberg96:
@bzwei @bronaghs Thoughts?

bzwei:
@djberg96 since configuration constructor already validate tenant ID, client ID/key, why don't we include validation of subscription ID there without making it an explicitly call for user?

djberg96:
@bzwei I'm ok with an up-front validation, but it does mean an extra http request they wouldn't otherwise need to make.

bzwei:
I am voting for this solution because it makes the validation behaviors the same for all user provided IDs.

djberg96:
@bzwei I updated the code to validate the subscription in the constructor and updated the specs.

PR #191, opened 2016-07-12
No longer need custom list_all method for AvailabilitySetService

djberg96:
It seems we can get all availability sets with the inherited list_all method now.


djberg96:
@bzwei Please review.

I think this just leaves TemplateDeployentService as the only service class still using it.

Issue #169, opened 2016-05-17
Use thread pooling for multi-threaded http requests


djberg96:
Jason Frey suggested that we use a thread pool of some sort to manage our threaded requests, both to manage the overall number of threads, and also to reduce the potential possibility of getting blocked by MS for making too many requests at once.

The candidates suggested by Jason are:

* https://github.com/ruby-concurrency/concurrent-ruby
* https://github.com/grosser/parallel

Whichever we choose, we need to decide what a good number is for the default (8?) and whether this should be a global setting, or a per-service class setting. Either way, I think this should be relegated to the 0.3.0 release.

@bzwei @bronaghs @roliveri @blomquisg @Fryguy - Feedback welcome.

PR #192, opened 2016-07-12
Use the parallel gem to cap our thread usage 

djberg96:
Addresses: https://github.com/ManageIQ/azure-armrest/issues/169

This replaces explicit `Thread.new` calls with `Parallel.each` calls, capping our thread pool at 10 threads per operation. It also adds a mutex for the `all_blobs` method (should have had one before) and updates the gemspec.

This is primarily for the benefit of the `list_in_all_groups` internal method, but is potentially useful elsewhere.


djberg96:
@bzwei Please review.

Issue #175, opened 2016-05-31
pretty_print failure on 2.3.x


djberg96:
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin15]

I'm not sure why it's not matching in 2.3.x, but here's the failure I'm seeing:

```
1) BaseModel inspection methods defines a pretty_print method when pp is available
     Failure/Error: expect(base.pretty_inspect).to match(expected)
     
       expected "#<Azure::Armrest::BaseModel:0x007f8fa35a96e8\n name=\"test\",\n array=[\"stuff\"],\n age=33>\n" to match /^#<Azure::Armrest::BaseModel:0x\h+\n name="test",\n age=33,\n array=\["stuff"\]>$/
       Diff:
       @@ -1,2 +1,5 @@
       -/^#<Azure::Armrest::BaseModel:0x\h+\n name="test",\n age=33,\n array=\["stuff"\]>$/
       +#<Azure::Armrest::BaseModel:0x007f8fa35a96e8
       + name="test",
       + array=["stuff"],
       + age=33>
       
     # ./spec/models/base_model_spec.rb:124:in `block (3 levels) in <top (required)>'

Finished in 0.86943 seconds (files took 0.72225 seconds to load)
305 examples, 1 failure
```

djberg96:
@Fryguy This was your baby. Any ideas?

Fryguy:
Looks like the order of the properties is different for some reason, so it's a spec-only failure.  I'll just have to make that test less brittle to that case.

djberg96:
Fixed in #193.

PR #193, opened 2016-07-12
Don't depend on key order for pretty_print spec.

djberg96:
This fixes a failing spec for 2.3.x where the key ordering is not reliable. It limits the regex check and then just looks for specific strings.

Addresses https://github.com/ManageIQ/azure-armrest/issues/175


djberg96:
@bzwei @Fryguy Please review.

PR #194, opened 2016-07-13
Update travis.yml to only test Ruby 2.2.5 or later

djberg96:
This limits our testing to Ruby 2.2.5 or 2.3.1.

Not only are older versions of Ruby mostly irrelevant, these are necessary to work with activesupport 5.x, and that's one of our dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 language: ruby
 rvm:
-- "2.0"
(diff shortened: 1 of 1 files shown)


djberg96:
The test failure is a known issue addressed in https://github.com/ManageIQ/azure-armrest/pull/193

PR #195, opened 2016-07-15
Add Hakiri Badge

chessbyte:


djberg96:
@chessbyte Seems it fails if our dependencies have failures. Not much we can do about those.

PR #196, opened 2016-07-20
Add a filter option to the list_all method.

djberg96:
I thought the filtering option for the `StorageAccountService#list_all_private_images` was quite handy, and would be generally useful for the `list_all` method.

This lets users do things like `vms.list_all(:location => 'westus')`, which I think is quite handy.

### DIFF
--- a/lib/azure/armrest/resource_group_based_service.rb
+++ b/lib/azure/armrest/resource_group_based_service.rb
@@ -23,11 +23,21 @@ module Azure
         JSON.parse(response)['value'].map { |hash| model_class.new(hash) }
       end
 
-      def list_all
+      # Use a single call to get all resources for the service. You may
+      # optionally provide a filter on various properties to limit the
+      # result set.
+      #
+      # Example:
+      #
+      #   vms = Azure::Armrest::VirtualMachineService.new(conf)
+      #   vms.list_all(:location => "eastus", :resource_group => "rg1")
+      #
+      def list_all(filter = {})
         url = build_url
         url = yield(url) || url if block_given?
         response = rest_get(url)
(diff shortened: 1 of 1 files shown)


djberg96:
@bzwei @bronaghs Look good?

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -23,11 +23,12 @@ def list(rgroup = configuration.resource_group)
         JSON.parse(response)['value'].map { |hash| model_class.new(hash) }
       end
 
-      def list_all
+      def list_all(filter = {})
```
bzwei: Maybe `filters`? Add comments/examples for the format of filters.

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -23,11 +23,12 @@ def list(rgroup = configuration.resource_group)
...
         url = build_url
         url = yield(url) || url if block_given?
         response = rest_get(url)
-        JSON.parse(response)['value'].map { |hash| model_class.new(hash) }
+        results = JSON.parse(response)['value'].map { |hash| model_class.new(hash) }
+        results.select { |obj| filter.all? { |k, v| obj.public_send(k) == v } }
```
bzwei: Check `filter` whether empty (better default to `nil` instead of `{}`) before `select` for performance sake.

djberg96: Ok, done.

lib/azure/armrest/resource_group_based_service.rb
djberg96: Eh, I tend to think of a filter as a single thing with multiple properties. Plus, it's shorter. ;)

I added some comments.

PR #198, opened 2016-07-21
Update gem dependencies to most recent versions.

djberg96:
Updating gem dependencies to see if we can eliminate Hakiri warnings.


djberg96:

This gets us past the Hakiri failures. Note that it does bump activesupport to 4.2.2 or later, so we are hosing any Rails 3 users. Hopefully, everyone has upgraded at this point.

PR #201, opened 2016-07-22
Add list_all_private_images to 0.2.x

djberg96:
I want to backport the `list_all_private_images` method into 0.2.x. This will help solve https://bugzilla.redhat.com/show_bug.cgi?id=1342649 without having to upgrade the azure-armrest gem, which requires more substantial changes.


bronaghs:
👍

PR #203, opened 2016-07-28
Update delete method to handle 204 response.

djberg96:
It turns out that the Azure REST API typically returns a 204 (empty body) on a DELETE operation for most resources if the resource does not exist instead of a 404 as you might expect. The net result is that you could call .delete on a resource multiple times and not realize that it didn't actually do anything. This PR specifically checks for a 204 and raises a ResourceNotFoundException if that's the response code.

More importantly, when calling delete we (eventually) want valid header information from the request that we can use to poll the resource because it's an asynchronous operation. If the resource does exist when we made the call to delete it, the header information will include a :location url attribute that we can use to poll the operation to see when it actually completes.

~~I'm also open to the approach of just returning true/false for the sake of backwards compatibility, in case anyone was relying on the old behavior.~~

Actually, we really, really need response header information, so I've updated the PR to return an actual ResponseHeader object that will contain the information we need.

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -194,6 +194,8 @@ module Azure
     class Sku < BaseModel; end
     class Usage < BaseModel; end
 
+    class ResponseHeaders < BaseModel; end
+
     class StorageAccount < BaseModel; end
     class StorageAccountKey < StorageAccount
       def key1; key_name == 'key1' ? value : nil; end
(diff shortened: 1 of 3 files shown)


djberg96:
@bronaghs @bzwei @blomquisg Please review.

bronaghs:
👍

bronaghs:
@djberg96 - will the MIQ retirement code need to be updated?

djberg96:
@bronaghs I wouldn't think so, since it would only fail if you try to delete a resource that's already been deleted, but I can double check. Do you have a link to it?

bronaghs:
this the vm delete i believe:
https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/azure/cloud_manager.rb#L98

PR #204, opened 2016-07-29
Add a :max_threads configuration option

djberg96:
This adds a :max_threads option with a default of 10. This value is used internally as the value for any calls to Parallel.

Because of thread safety issues caused by the interaction of the Parallel and VCR gems, the default is reduced to 1 in a VCR environment.

There was also a minor change made to the StorageAccount#all_blobs method so that we can control the maximum number of threads there as well.

### DIFF
--- a/lib/azure/armrest/configuration.rb
+++ b/lib/azure/armrest/configuration.rb
@@ -60,6 +60,9 @@ module Azure
       # Namespace providers, their resource types, locations and supported api-version strings.
       attr_reader :providers
 
+      # Maximum number of threads to use within methods that use Parallel for thread pooling.
+      attr_accessor :max_threads
+
       # Yields a new Azure::Armrest::Configuration objects. Note that you must
       # specify a client_id, client_key, tenant_id and subscription_id. All other
       # parameters are optional.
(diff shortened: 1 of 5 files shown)


djberg96:
@bronaghs @Fryguy Please review.

bronaghs:
@djberg96 
what were the threading issues fixed by this?

djberg96:
@bronaghs When using threads with the parallel gem in conjunction with the VCR gem, it can cause intermittent failures because VCR isn't entirely thread safe. I thought this was fixed with VCR 3, but apparently not.

What I'm doing here is basically what the Ovirt gem is already doing: 

https://github.com/ManageIQ/ovirt/blob/master/lib/ovirt/inventory.rb#L191-L198

lib/azure/armrest/configuration.rb
```diff
@@ -93,8 +96,12 @@ def initialize(args)
           :grant_type   => 'client_credentials',
           :proxy        => ENV['http_proxy'],
           :ssl_version  => 'TLSv1',
+          :max_threads  => 10
         }.merge(args.symbolize_keys)
 
+        # Avoid thread safety issues for VCR testing.
+        options[:max_threads] = 1 if defined?(VCR)
+
```
bronaghs: @djberg96 
Given that it is the MIQ Azure refresher_spec that is failing, where do we define VCR?

djberg96: We don't. If you don't like this approach I can remove it, but it does mean some stubbing in our refresher specs.

bronaghs: @djberg96 - If we do not set ```VCR``` how will we ever set the max_threads to 1?
Im trying to understand the logic, its not that i do not like it.

djberg96: In our testing environment, the VCR constant would be defined.

bronaghs: @djberg96 - OK thanks.

PR #205, opened 2016-08-03
Add poll and wait methods.

djberg96:
This adds poll and wait methods to the base ArmrestService class. This lets us take the result of an asynchronous option (e.g. delete), and poll the resulting url header for its operations status. 

The wait method is just a sleepy loop around the poll method that would let users essentially "block until complete" for async operations.

This is part of the continuing effort towards a "delete_associated" method.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -157,6 +157,45 @@ module Azure
         JSON.parse(resp.body)['value'].map{ |hash| Azure::Armrest::Tenant.new(hash) }
       end
 
+      # Poll a resource and return its current operations status. The
+      # +response+ argument should be a ResponseHeaders object. By
+      # default it will use the result of the :azure_asyncoperation
+      # header for its URL, but you may set :location instead.
+      #
+      # Alternatively, the first argument may be a plain URL string.
+      #
+      # This is meant to check the status of asynchronous operations,
+      # such as create or delete.
+      #
+      def poll(response, header = :azure_asyncoperation)
+        url = response.kind_of?(String) ? response : response.public_send(header)
+        JSON.parse(rest_get(url))['status']
+      end
+
+      # Wait for the given +response+ to return a status of 'Succeeded', up
+      # to a maximum of +max_time+ seconds, and return the operations status.
+      # The first argument may be a ResponseHeaders object, or a plain URL string.
+      #
+      # Internally this will poll the response header every 10 seconds (or
+      # whatever the :retry_after header is set to), up to a maximum of 60
+      # seconds by default.
+      #
+      # For most resources the +max_time+ argument should be more than sufficient.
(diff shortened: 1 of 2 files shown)


djberg96:

lib/azure/armrest/armrest_service.rb
```diff
@@ -157,6 +157,45 @@ def tenants
...
+        while (status = poll(response)) != 'Succeeded'
+          total_time += sleep_time
+          break if total_time >= max_time
+          sleep sleep_time
+        end
+
+        status
+      end
+
       class << self
```
bronaghs: @djberg96 
Where do you read :location?

djberg96: I suppose I could add that as an optional 3rd argument that is forwarded to the `poll` method.

bronaghs: whichever you choose should be consistent with specs

djberg96: @bronaghs Ok, updated the method and the specs.

lib/azure/armrest/armrest_service.rb
```diff
@@ -157,6 +157,48 @@ def tenants
...
+      # Certain resources, such as virtual machines, could take longer.
+      #
+      # The optional +header+ argument is forwarded to the poll method and
+      # determines which header is used for polling the operation.
+      #
+      def wait(response, max_time = 60, header = :azure_asyncoperation)
+        sleep_time = response.respond_to?(:retry_after) ? response.retry_after.to_i : 10
+        total_time = 0
+
+        while (status = poll(response, header)) != 'Succeeded'
```
bzwei: Are `:azure_syncoperation` and `:location` the only choices? If yes, then don't make it an argument. You can test internally.
If `poll` is always paired with `wait`, then don't offer the first argument of `poll` alternatively a string, but always a `ResponseHeaders`

djberg96: Yes, they are the only choices. I guess I was trying to be flexible since I found this Python lib:

http://msrestazure.readthedocs.io/en/latest/msrestazure.html

Which has `get_status_from_location`, `get_status_from_async` and `get_status_from_resource` methods.

As to how useful those are in practice, I guess it's hard to say. My suspicion is that we should always prefer :azure_asyncoperation first, which is what I originally did, then changed it.

bronaghs:
@bzwei - look good?

bzwei:
@djberg96 @bronaghs Please see my earlier comment. I feel we do not need to introduce too many flexibilities, so no need to make `header` an argument and `response` possibly two types.

djberg96:
@bzwei I modified it to just be :azure_asyncoperation by default, and anything that responds to .to_s as a fallback.

That way users could use a plain string, a URI object, or anything that responds to to_s. So, still some flexibility but without the extra arguments.

PR #207, opened 2016-08-04
Add RouteService and RouteTableService classes.

djberg96:
Just rounding out the networking bits, I've added support for RouteService and RouteTableService classes, as well as correponding specs.


bzwei:
@djberg96 are we going to collect them in refresh or other places?

djberg96:
@bzwei Perhaps. I wasn't sure if this would be useful with the upcoming push for load balancer support. But, in any case, other users might find it useful. :)

djberg96:
@bzwei Just realized I never pushed up the specs. Added.

PR #208, opened 2016-08-06
Fix and refactor the StorageAccount#blobs method.

djberg96:
The current code for StorageAccount#blobs has two problems.

The first one is that it doesn't support all of the possible options that are available. The third argument allows for the addition of snapshots, but that's it. There are many more possible options users should be able to choose them - prefix, delimiter, timeout, include, maxresults and marker. Furthermore, in addition to snapshots, there are also metadata, uncommittedblobs and copy blob options that users can select as part of the `:include` option.

Secondly, it turns out that in some cases (3 out of 104 cases in local testing), the code requires multiple calls to get the blob information. The first call returns only partial information and sets a `NextMarker` element token. We must then make a second call with this token as part of the URI. For more info see:

https://msdn.microsoft.com/en-us/library/azure/hh452233.aspx

This is similar to a skip token, except that there doesn't seem to be any direct correlation between `maxresults` and `NextMarker` that I can see. In my 3 local examples, none had more than a single blob or container, so I do not understand why it gets set, but there it is.

*Update: I found a snippet from "Developing Cloud Applications with Windows Azure Storage" that indicates this can happen if the underlying data was moved to a different underlying storage node. The marker tells it to get the data from the new node.*

Currently the code returns an empty array for these instances, which is incorrect, and could potentially impact private image listing as well.

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -119,14 +119,57 @@ module Azure
       # If no key is provided, it is assumed that the StorageAccount object
       # includes the key1 property.
       #
-      def containers(key = nil)
+      # # The following options are supported:
+      #
+      # * prefix
+      # * delimiter
+      # * maxresults
+      # * include
+      # * timeout
+      #
+      # By default Azure uses a value of 5000 for :maxresults.
+      #
+      # If the :include option is specified, it should contain an array of
+      # one element: metadata. More options may be added by Microsoft
+      # at a later date.
+      #
+      # Example:
+      #
+      #   sas  = Azure::Armrest::StorageAccountService.new(conf)
+      #   key  = sas.list_account_keys['key1']
+      #   acct = sas.get('your_storage_account', 'your_resource_group')
+      #
+      #   p acct.containers(key)
+      #   p acct.containers(key, :include => ['metadata'])
+      #   p acct.containers(key, :maxresults => 1)
+      #
+      # In cases where a NextMarker element is found in the original response,
+      # another call will automatically be made with the marker value included
+      # in the URL so that you don't have to perform such a step manually.
+      #
+      def containers(key = nil, options = {})
         key ||= properties.key1
 
-        response = blob_response(key, "comp=list")
+        query = "comp=list"
+        options.each { |okey, ovalue| query += "&#{okey}=#{[ovalue].flatten.join(',')}" }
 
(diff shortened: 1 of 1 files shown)


djberg96:

bzwei:
@djberg96 Looks good. What's the impact for backward compatibility?

djberg96:
@bzwei The only issue would be that optional argument to the blobs argument. However, we don't use it within CloudForms, and I'm guessing most people in the wild weren't using it either.

Also, I've updated the PR, refactored the `blobs` method a bit, and also added options and `NextMarker` support for the `containers` method as well.

lib/azure/armrest/model/storage_account.rb
```diff
@@ -119,14 +119,57 @@ def table_data(name, key = nil, options = {})
...
+      #
+      #   p acct.containers(key)
+      #   p acct.containers(key, :include => ['metadata'])
+      #   p acct.containers(key, :maxresults => 1)
+      #
+      # In cases where a NextMarker element is found in the original response,
+      # another call will automatically be made with the marker value included
+      # in the URL so that you don't have to perform such a step manually.
+      #
+      def containers(key = nil, options = {})
```
bzwei: are there any properties in addition to container and blobs that requires such refactoring?

lib/azure/armrest/model/storage_account.rb
```diff
@@ -119,14 +119,57 @@ def table_data(name, key = nil, options = {})
...
+        response = blob_response(key, query)
 
-        Nokogiri::XML(response.body).xpath('//Containers/Container').map do |element|
+        doc = Nokogiri::XML(response.body)
+
+        results = doc.xpath('//Containers/Container').collect do |element|
           Container.new(Hash.from_xml(element.to_s)['Container'])
         end
+
+        doc.xpath('//NextMarker').each do |xmarker|
```
bzwei: is it possible to pull this code into a reusable method? the logic is the same for containers and blobs

lib/azure/armrest/model/storage_account.rb
djberg96: There might be. My assumption is that these were added after we originally created this method. If there are, I'll do a separate PR.

lib/azure/armrest/model/storage_account.rb
djberg96: I suppose I can do that. :)

djberg96: @bzwei Done.

PR #209, opened 2016-08-10
Set proxy before calling validate_subscription.

djberg96:
At the moment a proxy is not getting set prior to the call to validate_subscription, which can lead to failures, especially as rest-client still uses URI instead of Addressable.

This mistake effectively prevents Cloudforms from working with authenticating proxies at the moment.


djberg96:
@bzwei Please review.

bzwei:
Looks good

PR #210, opened 2016-08-16
Add comments for exception classes, and update to_s method. 

djberg96:
This adds comments to the `Armrest::Exception` and `Armrest::ApiException` classes in exception.rb.

I also added a separate set of specs for the exceptions themselves.

The only real code change I made was to the `ApiException#to_s` method so that it includes the cause (if present), which should aid in debugging when one of our wrapper classes doesn't catch the underlying RestClient exception.

### DIFF
--- a/lib/azure/armrest/exception.rb
+++ b/lib/azure/armrest/exception.rb
@@ -4,15 +4,30 @@ module Azure
       attr_accessor :cause
       attr_writer :message
 
+      # Create a new Armrest::Exception object. The +message+ should be an
+      # error string, while +cause_exception+ is typically set to the
+      # raw RestClient exception.
+      #
+      # You will not typically use this object directly.
+      #
       def initialize(message = nil, cause_exception = nil)
         @message = message
         @cause = cause_exception
       end
 
+      # The stringified version (message) of the exception.
+      #
       def to_s
-        message
+        if cause
+          "#{message} (cause: #{cause})"
+        else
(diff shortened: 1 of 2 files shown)


djberg96:
@bzwei Please review.

lib/azure/armrest/exception.rb
```diff
@@ -21,16 +32,28 @@ def message
...
         @code = code
         super(message, cause_exception)
       end
 
+      # A stringified version of the error. If self is a plain ApiException,
+      # then the cause is included to aid in debugging.
+      #
       def to_s
-        "[#{code}] #{message}"
+        if self.class == ApiException && cause
```
bzwei: Why print cause only for plain ApiException? The cause can be always included in the base class's to_s, something like
```
"#{message}, cause: #{cause}"
```
Then here it can be simply
```
"[#{code}] #{super}"
```

djberg96: Because you know the error based on the name, e.g. a TooManyRequests exception already tells me the cause - too many requests. The cause then becomes redundant. But, with ApiException it isn't immediately clear.

bzwei: It may not be bad to always print the cause. The cause_exception's `to_s` may give more detailed error message.

djberg96: Alright, I've updated the PR.

PR #214, opened 2016-09-16
Add response_code accessor and use it internally.

djberg96:
This adds a response_code accessor to our base model, sets it internally for our methods, and uses it within the poll method.

It also modifies the create method to return a response header object instead of a regular object since it's an asynchronous operation, and makes it consistent with the delete method.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -166,7 +166,7 @@ module Azure
       # such as create or delete.
       #
       def poll(response)
-        return 'Succeeded' if [200,201].include?(response.response_code)
+        return 'Succeeded' if [200, 201].include?(response.response_code)
         url = response.try(:azure_asyncoperation) || response.try(:location)
         JSON.parse(rest_get(url))['status']
       end
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -166,7 +166,8 @@ module Azure
       # such as create or delete.
       #
       def poll(response)
-        url = response.respond_to?(:azure_asyncoperation) ? response.azure_asyncoperation : response.to_s
+        return 'Succeeded' if [200,201].include?(response.response_code)
+        url = response.try(:azure_asyncoperation) || response.try(:location)
         JSON.parse(rest_get(url))['status']
       end
 
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -24,6 +24,7 @@ module Azure
       attr_hash :tags
 
       attr_accessor :response_headers
+      attr_accessor :response_code
 
       # Constructs and returns a new JSON wrapper class. Pass in a plain
       # JSON string and it will automatically give you accessor methods
@@ -189,7 +190,9 @@ module Azure
     class Sku < BaseModel; end
     class Usage < BaseModel; end
 
-    class ResponseHeaders < BaseModel; end
+    class ResponseHeaders < BaseModel
+      undef_method :response_headers
+    end
 
     class StorageAccount < BaseModel; end
     class StorageAccountKey < StorageAccount
--- a/spec/template_deployment_service_spec.rb
+++ b/spec/template_deployment_service_spec.rb
@@ -8,11 +8,8 @@ require 'spec_helper'
 describe "TemplateDeploymentService" do
   before do
     setup_params
-
-    class String
-      def body; self; end
-      def headers; {}; end
-    end
+    allow_any_instance_of(String).to receive(:headers).and_return({})
+    allow_any_instance_of(String).to receive(:code).and_return(200)
   end
 
   let(:tds) { Azure::Armrest::TemplateDeploymentService.new(@conf) }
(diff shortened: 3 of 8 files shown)


djberg96:
@bzwei @bronaghs Please take a look at this and let me know what you think. The short version is that we need the response code, which is not included in the headers or the body (is it?). For the `poll` method, as per the docs and Boris' email, we can assume success on a 200 or 201.

@bronaghs This (or something like it) is necessary before I can continue with deleting associated resources.

djberg96:
@bzwei I'm having trouble with one test yet, from TemplateDeployments:

```
1) TemplateDeploymentService instance methods defines a get method
     Failure/Error: @hash = JSON.parse(json)
     
     TypeError:
       no implicit conversion of RSpec::ExampleGroups::TemplateDeploymentService::InstanceMethods into String
     # ./lib/azure/armrest/model/base_model.rb:62:in `initialize'
     # ./lib/azure/armrest/resource_group_based_service.rb:78:in `new'
     # ./lib/azure/armrest/resource_group_based_service.rb:78:in `get'
     # ./spec/template_deployment_service_spec.rb:70:in `block (3 levels) in <top (required)>'
```

Can you take a look? I'm not sure how to fix that.

spec/template_deployment_service_spec.rb
```diff
@@ -8,11 +8,9 @@
 describe "TemplateDeploymentService" do
   before do
     setup_params
-
-    class String
-      def body; self; end
-      def headers; {}; end
-    end
+    allow_any_instance_of(String).to receive(:body).and_return(self)
```
bzwei: returning `self` caused the spec test failure because `self` is `RSpec::ExampleGroups::...InstanceMethods`. A string is expected from the `body` method.

djberg96: @bzwei Thanks, fixed.

bronaghs:
@djberg96 - We no longer create the object itself eg a VM? We need this object for instance provisioning...
https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/azure/cloud_manager/provision/cloning.rb#L143

djberg96:
@bronaghs We still create the object, it's just that the create method returns the response headers instead of internally making another get call.

I suppose I could keep it backwards compatible, and just attach the response info to the object itself.

bronaghs:
@djberg96 - I realize we are still creating the object, what I meant was we need the object itself to be returned. Can you point me to where it is being returned?

bronaghs:
@djberg96 - I mistyped my comment above which explains your response, I should have said:
"We no longer ***return*** the object itself eg a VM? We need this object for instance provisioning..."

bzwei:
I agree with @djberg96 The create method should always return the object it creates.

djberg96:
@bzwei @bronaghs Ok, the :create method returns an object as it did before. I've removed the WIP label, as I think it's ready to go.

bronaghs:
LGTM, @bzwei ?

lib/azure/armrest/armrest_service.rb
```diff
@@ -166,7 +166,8 @@ def tenants
       # such as create or delete.
       #
       def poll(response)
-        url = response.respond_to?(:azure_asyncoperation) ? response.azure_asyncoperation : response.to_s
+        return 'Succeeded' if [200, 201].include?(response.response_code)
```
bzwei: For my curiosity, what is the response_code if the aync operation is not finished yet?
Also, for the case the response contains `azure_asyncoperation` or `location`, what is the return code when it is still ongoing or finished?

djberg96: If it's not finished yet it returns a 202.

https://httpstatuses.com/202

lib/azure/armrest/model/base_model.rb
```diff
@@ -189,7 +190,9 @@ class ResourceProvider < BaseModel; end
     class Sku < BaseModel; end
     class Usage < BaseModel; end
 
-    class ResponseHeaders < BaseModel; end
+    class ResponseHeaders < BaseModel
+      undef_method :response_headers
```
bzwei: undefine a method in a subclass that exists in base class is not an object orient way. Is there other way around?

djberg96: @bzwei I just thought it was inappropriate for a ResponseHeaders object to have a :response_headers method. It was actually causing some confusion for me in development.

If there's a nicer way, I'm listening.

bzwei: @djberg96 `ResponseHeaders` does not need to be based on the `BaseModel`.

djberg96: @bzwei Well, I need to be able to create a model object from the response.headers output from RestClient. What do you suggest then?

djberg96: Not really, since I need the automatic method generation.

I think down the road we should actually create an ExtendedModel subclass that adds additional accessors like :response_headers, plus any other tweaks we need. Most models would then subclass ExtendedModel, while something like ResponseHeaders would just subclass BaseModel. But, I would ask that be a separate PR.

bzwei:
OK. Subject to further PR refactoring about the base class

Issue #215, opened 2016-09-19
The list methods don't paginate


djberg96:
At the moment neither the `list` nor the `list_all` methods paginate, which would result in truncated results. They need to handle skip tokens.

djberg96:
Fixed by #216.

PR #216, opened 2016-09-22
Update the list and list_all methods to support pagination.

djberg96:
This PR modifies the list and list_all methods in the ResourceGroupBasedService to support pagination.

Addresses https://github.com/ManageIQ/azure-armrest/issues/215.

~~~This required the addition of an accessor to the ArmrestCollection object called continuation_url, since it gives us a full URL instead of just a token.~~~

I also had to implement a new method that would not automatically encode the URL. This is because the URL is "pre-encoded" and encoding it again caused incorrect (and strange) results.

### DIFF
--- a/lib/azure/armrest/resource_group_based_service.rb
+++ b/lib/azure/armrest/resource_group_based_service.rb
@@ -46,7 +46,7 @@ module Azure
         url = yield(url) || url if block_given?
         response = rest_get(url)
 
-        Azure::Armrest::ArmrestCollection.create_from_response(response, model_class)
+        get_all_results(response)
       end
 
       # Use a single call to get all resources for the service. You may
@@ -63,7 +63,7 @@ module Azure
         url = yield(url) || url if block_given?
 
         response = rest_get(url)
-        results  = Azure::Armrest::ArmrestCollection.create_from_response(response, model_class)
+        results  = get_all_results(response)
 
         filter.empty? ? results : results.select { |obj| filter.all? { |k, v| obj.public_send(k) == v } }
       end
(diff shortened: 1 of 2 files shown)


djberg96:

djberg96:
Actually, I'll modify this so that it's backwards compatible with the 0.2.x branch so that it can be merged. We can update it later in a separate PR for 0.3.x.

Marked as WIP for now...

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -65,6 +74,13 @@ def list_all(filter = {})
...
         results  = Azure::Armrest::ArmrestCollection.create_from_response(response, model_class)
 
+        while results.continuation_url
+          response = rest_get_without_encoding(results.continuation_url)
+          more = Azure::Armrest::ArmrestCollection.create_from_response(response, model_class)
+          results.concat(more)
+          results.continuation_url = more.continuation_url
+        end
+
         filter.empty? ? results : results.select { |obj| filter.all? { |k, v| obj.public_send(k) == v } }
```
bronaghs: @djberg96 - maybe move into a method since it is duplicated?

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -46,7 +46,16 @@ def list(rgroup = configuration.resource_group)
...
+
+        while results.continuation_url
+          response = rest_get_without_encoding(results.continuation_url)
+          more = Azure::Armrest::ArmrestCollection.create_from_response(response, model_class)
+          results.concat(more)
+          results.continuation_url = more.continuation_url
+        end
+
+        results
       end
```
bronaghs: @djberg96 - I would have created a separate local variable called eg ```continuation_url``` and set/updated it as needed; it feels uncomfortable to me to overwrite the url in the original ```results``` array. This is only a minor style comment though, feel free to ignore or take on board.

djberg96: @bronaghs Maybe it's not even worth exposing this as an accessor to begin with. I'll update the PR to remove it and just handle it internally for now since it's only used in 2 methods at the moment. We could always revisit.

bronaghs:
@djberg96 - Looks great 👍

PR #217, opened 2016-09-22
Update delete method to handle 204 response.

djberg96:
Backports https://github.com/ManageIQ/azure-armrest/pull/203

Update delete spec for TemplateDeploymentService.
Update delete method to return ResponseHeaders object.

### DIFF
--- a/lib/azure/armrest/resource_group_based_service.rb
+++ b/lib/azure/armrest/resource_group_based_service.rb
@@ -40,14 +40,28 @@ module Azure
         model_class.new(response)
       end
 
+      # Delete the resource with the given +name+ for the provided +resource_group+,
+      # or the resource group specified in your original configuration object. If
+      # successful, returns a ResponseHeaders object.
+      #
+      # If the delete operation returns a 204 (no body), which is what the Azure
+      # REST API typically returns if the resource is not found, it is treated
+      # as an error and a ResourceNotFoundException is raised.
+      #
       def delete(name, rgroup = configuration.resource_group)
         validate_resource_group(rgroup)
         validate_resource(name)
 
         url = build_url(rgroup, name)
         url = yield(url) || url if block_given?
-        rest_delete(url)
-        nil
+        response = rest_delete(url)
+
+        if response.code == 204
+          msg = "#{self.class} resource #{rgroup}/#{name} not found"
+          raise Azure::Armrest::ResourceNotFoundException.new(response.code, msg, response)
+        end
+
+        Azure::Armrest::ResponseHeaders.new(response.headers)
       end
 
       private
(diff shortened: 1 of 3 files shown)


djberg96:
@bronaghs @bzwei This won't cause any problems for you, will it?

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -40,14 +40,28 @@ def get(name, rgroup = configuration.resource_group)
...
-        rest_delete(url)
-        nil
+        response = rest_delete(url)
+
+        if response.code == 204
+          msg = "#{self.class} resource #{rgroup}/#{name} not found"
+          raise Azure::Armrest::ResourceNotFoundException.new(response.code, msg, response)
+        end
+
+        Azure::Armrest::ResponseHeaders.new(response.headers)
```
bzwei: Why do we ever need to return a `ResponseHeaders`? Although the backend we interact with the provider through REST API, the application (user) should not know/care it is a HTTP call.

djberg96: The response headers contain the URL's necessary to check the status of the operation since it's asynchronous. We use these for polling.

bzwei: I read the original PR #203 and understood the use of response header to pull status of the asynchronous call. 
If the only use is for asynchronous call, can we create a new class to wrap the header more properly, and provide one only method to report the status of the asynchronous call?

djberg96: What do you mean "a new class to wrap the header more properly"? What more do we need than what I've got there?

bzwei: The `delete` method could return an instance of a new class, say `AsyncHandle`. The class has one method `AsyncHandle#status`. The class does not need to inherit from `BaseModel`.

djberg96: @bzwei That seems more complex and less useful than what I'm doing. I'm basically just returning what the delete request returns, as an object.

bzwei: @djberg96 That's the exact point. What the use of the simple wrapping of the response header? The caller should not know/care it is a HTTP response. What the caller cares is how to monitor the status if it is an asynchronous call.

djberg96: @bzwei And in order to monitor the status, we need the response headers. They could also use the response headers to check other information, such as ratelimit info.

bzwei:
I don't see it might break anything as the return value for `delete` is normally not checked.

djberg96:
@bzwei That's what I thought, but wanted to double check, thanks.

PR #165, opened 2016-05-09
Add support for PP

Fryguy:
Because the `inspect` method was overridden, Ruby's PP module (and by extension `pry`'s pretty printer) would not pretty print any objects properly, instead leaving them on a single line which is hard to use.  This PR allows PP to work properly.

Before:

```ruby
$ bundle exec irb -Ilib -r"azure-armrest" -rpp
irb(main):001:0> o = Azure::Armrest::BaseModel.new({:name => 'test', :age => 33, :array => ["stuff"]}.to_json)
=> <Azure::Armrest::BaseModel name="test", age=33, array=["stuff"]>
irb(main):002:0> pp o
<Azure::Armrest::BaseModel name="test", age=33, array=["stuff"]>
=> <Azure::Armrest::BaseModel name="test", age=33, array=["stuff"]>
```

After

```ruby
$ bundle exec irb -Ilib -r"azure-armrest" -rpp
irb(main):001:0> o = Azure::Armrest::BaseModel.new({:name => 'test', :age => 33, :array => ["stuff"]}.to_json)
=> #<Azure::Armrest::BaseModel:0x007fe3b5282d28 name="test", age=33, array=["stuff"]>
irb(main):002:0> pp o
#<Azure::Armrest::BaseModel:0x007fe3b5282d28
 name="test",
 age=33,
 array=["stuff"]>
=> #<Azure::Armrest::BaseModel:0x007fe3b5282d28 name="test", age=33, array=["stuff"]>
```

PR #219, opened 2016-09-30
Replace custom inspect with the one provided by pp.

Fryguy:
Since we are already implementing the pretty_print method, we can rely
on the built-in pretty_print_inspect method, whose purpose is to
provide an inspect method directly from a custom pretty_print method.

This is a follow on from #165 , now I've learned a bit more about PP and found out about this pretty_print_inspect method.  By reusing that method, we can eliminate our custom inspect method entirely.

@djberg96 In the original PR, you were concerned it would affect AP, which it didn't.  Can you also verify it doesn't affect AP here as well?

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -1,4 +1,5 @@
 require 'active_support/core_ext/string/inflections'
+require 'pp'
 
 module Azure
   module Armrest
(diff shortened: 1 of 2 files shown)


djberg96:
@Fryguy It doesn't seem to affect awesome_print, and I'm always in favor of removing unused code. :)

Issue #218, opened 2016-09-29
Need a better way to wrap RestClient exceptions


djberg96:
At the moment we have a series of pre-defined exception classes in exception.rb. Unfortunately, this means that if RestClient raises an error that we don't define, users get a generic ApiException. This is not usually very helpful in debugging.

What I would like to see is some sort of code that automatically transforms the RestClient exceptions into Armrest exceptions in a more dynamic way so that we don't have to add them manually.

djberg96:
Fixed by #220, closing.

PR #220, opened 2016-09-30
Dynamically define our own exception classes based on the RestClient

djberg96:
Addresses https://github.com/ManageIQ/azure-armrest/issues/218

At the moment we're hard coding our exception classes in exception.rb, as well as rescuing and re-raising them individually in armrest_service.rb. While this works, whenever RestClient raises an error that we don't explicitly define, it defaults to ApiException.

Defaulting to ApiException is unfriendly for debugging purposes and leads to unnecessary guesswork, and there are a lot of exceptions that we don't define ourselves. This dynamically defines an exception class for each RestClient exception, and dynamically determines which one to use if an error is raised. The only ones we need to define ourselves are the ones that aren't defined by RestClient.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -230,34 +230,30 @@ module Azure
           rest_execute(options, :head)
         end
 
-        def raise_api_exception(e)
+        def raise_api_exception(err)
           begin
-            response = JSON.parse(e.http_body)
-            code = response['error']['code']
-            message = response['error']['message']
+            response = JSON.parse(err.http_body)
+            code     = response['error']['code']
+            message  = response['error']['message']
           rescue
-            message = e.http_body
+            code = err.try(:http_code) || err.try(:code)
+            message = err.try(:http_body) || err.try(:message)
           end
-          message = e.http_body unless message
 
-          exception_type = case e
-                           when RestClient::NotFound
-                             ResourceNotFoundException
-                           when RestClient::BadRequest
-                             BadRequestException
-                           when RestClient::GatewayTimeout
-                             GatewayTimeoutException
-                           when RestClient::BadGateway
-                             BadGatewayException
-                           when RestClient::Unauthorized, RestClient::Forbidden
-                             UnauthorizedException
-                           when RestClient::TooManyRequests
-                             TooManyRequestsException
-                           else
-                             ApiException
-                           end
+          exception_type = Azure::Armrest::EXCEPTION_MAP[err.http_code]
 
-          raise exception_type.new(code, message, e)
+          # If this is an exception that doesn't map directly to an HTTP code
+          # then parse it the exception class name and re-raise it as our own.
+          if exception_type.nil?
+            begin
+              klass = "Azure::Armrest::" + err.class.to_s.split("::").last + "Exception"
+              exception_type = const_get(klass)
+            rescue NameError
+              exception_type = Azure::Armrest::ApiException
+            end
+          end
+
+          raise exception_type.new(code, message, err)
         end
       end
 
--- a/lib/azure/armrest/exception.rb
+++ b/lib/azure/armrest/exception.rb
@@ -52,18 +52,100 @@ module Azure
       end
     end
 
-    # A list of predefined exceptions that we wrap around RestClient exceptions.
-
-    class ResourceNotFoundException < ApiException; end
-
-    class BadRequestException < ApiException; end
-
-    class UnauthorizedException < ApiException; end
-
+    # Rewrapped HTTP errors
     class BadGatewayException < ApiException; end
-
+    class BadRequestException < ApiException; end
+    class BandwidthLimitExceededException < ApiException; end
+    class BlockedByWindowsParentalControlsException < ApiException; end
+    class ConflictException < ApiException; end
+    class ExpectationFailedException < ApiException; end
+    class FailedDependencyException < ApiException; end
+    class ForbiddenException < ApiException; end
     class GatewayTimeoutException < ApiException; end
-
+    class GoneException < ApiException; end
+    class HTTPVersionNotSupportedException < ApiException; end
+    class ImATeapotException < ApiException; end
+    class InsufficientStorageException < ApiException; end
+    class InternalServerErrorException < ApiException; end
+    class LengthRequiredException < ApiException; end
+    class LockedException < ApiException; end
+    class LoopDetectedException < ApiException; end
+    class MethodNotAllowedException < ApiException; end
+    class NetworkAuthenticationRequiredException < ApiException; end
+    class NotAcceptableException < ApiException; end
+    class NotExtendedException < ApiException; end
+    class NotFoundException < ApiException; end
+    class NotImplementedException < ApiException; end
+    class PayloadTooLargeException < ApiException; end
+    class PaymentRequiredException < ApiException; end
+    class PreconditionFailedException < ApiException; end
+    class PreconditionRequiredException < ApiException; end
+    class ProxyAuthenticationRequiredException < ApiException; end
+    class RangeNotSatisfiableException < ApiException; end
+    class RequestHeaderFieldsTooLargeException < ApiException; end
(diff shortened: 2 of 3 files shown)


djberg96:
@bzwei @bronaghs @Fryguy Please review and make sure this is a sound approach.

lib/azure/armrest/exception.rb
```diff
@@ -52,18 +52,16 @@ def to_s
       end
     end
 
-    # A list of predefined exceptions that we wrap around RestClient exceptions.
+    # Redefine the RestClient exception classes as our own.
+    RestClient::Exceptions::EXCEPTIONS_MAP.values.each do |klass|
+      name = klass.to_s.split("::").last
+      if name != 'Exception'
+        name += 'Exception' unless name.include?('Exception')
+        Object.module_eval("class Azure::Armrest::#{name} < Azure::Armrest::ApiException; end")
```
chrisarcand: 😕 

Can you utilize something like `MyError = Class.new(ErrorImInheritingFrom)` instead of `Object.module_eval`? (The dynamic nature of this has some small caveats you'll need to consider but I think with a little research it's possible)

djberg96: @chrisarcand If there's a way, I don't know what it is. I did update the PR to just use class_eval instead of Object.class_eval since we want to make sure we're evaluating in its current context. The tests confirm it.

spec/exception_spec.rb
```diff
@@ -75,6 +75,9 @@
       expect(Azure::Armrest::BadGatewayException).to_not be_nil
       expect(Azure::Armrest::GatewayTimeoutException).to_not be_nil
       expect(Azure::Armrest::TooManyRequestsException).to_not be_nil
+      expect(Azure::Armrest::PayloadTooLargeException).to_not be_nil
+      expect(Azure::Armrest::ServiceUnavailableException).to_not be_nil
+      expect(Azure::Armrest::LoopDetectedException).to_not be_nil
     end
```
bronaghs: @djberg96 - Can you add more meaty specs?

bronaghs:
@djberg96  Other than my specs comment I think this is great.

spec/exception_spec.rb
djberg96: @bronaghs I suppose I could add something, though the mere existence of those classes tells us that the generation happened successfully at least.

chrisarcand: A more semantically truthy (and nicer error messages) way would be `expect(const_defined?(Azure::Armrest::Exception)).to be_truthy` (note you can also do `expect(Azure::Armrest::Exception).to be_const_defined` but I dislike the grammar bastardization here intensely...)

djberg96: @chrisarcand You sure that's supposed to work? Blows up on me.

chrisarcand: How about `Azure::Armrest.const_defined?(Thing)` ?

bzwei:
@djberg96 The purpose that we originally explicitly define a set of subtypes of `Azure::Amrest::Exception` is to decouple the dependency on `RestClient`. If we ever decide to replace `RestClient`, the applications that use `armrest` gem do not need to make any change.

With this refactoring of auto-naming the exceptions it strongly couples with `RestClient`. Immediately manageiq, the current major consumer already needs to make corresponding change.

I opt to manually add a new `Azure::Amreset::Exception` type as soon as we find it is necessary, although it may look like very much its `RestClient` peer, they are not bound each other. 

For the unnamed ones that get converted to `ApiException`, although the naming itself is not very helpful, we should make it traceable so no important information is missed.

chrisarcand:
> With this refactoring of auto-naming the exceptions it strongly couples with RestClient. Immediately manageiq, the current major consumer already needs to make corresponding change.

That's not necessarily true, is it? When/if this gem's dependency changes, it can still remain backward compatible with these error names. Still, you're right, it does couple us to support whatever RestClient adds after the fact.

> For the unnamed ones that get converted to ApiException, although the naming itself is not very helpful, we should make it traceable so no important information is missed.

That was my other thought as well, we could just make it more helpful by adding the internal error within the ApiException message. `error: Azure::Armrest::ApiException: The underlying exception was: RestClient::SomeNewException, <trace>`

djberg96:
@bzwei @chrisarcand We're already tightly coupled in exception.rb (look at our original case/when clause, for example) as well as elsewhere in our code. I don't see this code making us more tightly coupled in any significant way.

bzwei:
@djberg96 Lookup and translate is OK. This is the place we connect RestClient and Armrest gem, also the place we modify to replace RestClient. 

Auto-naming after RestClient is no good because if RestClient ever renames an exception, it causes a renaming of Armrest exception which is not desired. 

If RestClient is replaced with another gem that uses another set of exceptions, it also causes renaming of Armrest exceptions, again, not desired.

djberg96:
@bzwei Ok, I've revamped the PR so that it defines exception classes manually. In order to decouple the code from RestClient even more I've setup a class map based on the HTTP error code, rather than the big case/when statement we currently have.

I also added some more specs. However, one of them is failing and I don't understand why. The message from rspec is not helpful to me. Help wanted.

lib/azure/armrest/exception.rb
```diff
@@ -52,18 +52,100 @@ def to_s
...
-    class GatewayTimeoutException < ApiException; end
+    # Custom errors or other wrapped exceptions
+    class ResourceNotFoundException < ApiException; end
+    class TimeoutException < RequestTimeoutException; end
+    class OpenTimeoutException < TimeoutException; end
+    class ReadTimeoutException < TimeoutException; end
 
-    class TooManyRequestsException < ApiException; end
+    # Map HTTP error codes to our exception classes
+    EXCEPTION_MAP = {
```
bzwei: 💯

lib/azure/armrest/armrest_service.rb
```diff
@@ -230,34 +230,30 @@ def rest_head(options)
...
-
-          raise exception_type.new(code, message, e)
+
+          exception_type = Azure::Armrest::EXCEPTION_MAP[err.http_code]
+
+          # If this is an exception that doesn't map directly to an HTTP code
+          # then parse it the exception class name and re-raise it as our own.
+          if exception_type.nil?
+            begin
+              klass = "Azure::Armrest::" + err.class.to_s.split("::").last + "Exception"
```
bzwei: We can do this way for now. The proper way would be using another map like how you do with the `http_code`.

bzwei:
@djberg96  Can you restart the build and make it green?

djberg96:
@bzwei One of the tests is failing but I don't understand it.

spec/exception_spec.rb
```diff
@@ -68,13 +68,40 @@
...
+      expect(Azure::Armrest.const_defined?(:GatewayTimeoutException)).to be_truthy
+      expect(Azure::Armrest.const_defined?(:TooManyRequestsException)).to be_truthy
+      expect(Azure::Armrest.const_defined?(:PayloadTooLargeException)).to be_truthy
+      expect(Azure::Armrest.const_defined?(:ServiceUnavailableException)).to be_truthy
+      expect(Azure::Armrest.const_defined?(:LoopDetectedException)).to be_truthy
+    end
+
+    it "re-raises the expected error class" do
+      expect(armrest_service.send(:raise_api_exception, badrequest)).to raise_error(Azure::Armrest::BadRequestException)
+      expect(armrest_service.send(:raise_api_exception, timeout)).to raise_error(Azure::Armrest::TimeoutException)
```
chrisarcand:
```ruby
expect { armrest_service.send(:raise_api_exception, badrequest) }
  .to raise_error(Azure::Armrest::BadRequestException)
expect { armrest_service.send(:raise_api_exception, timeout) }
  .to raise_error(Azure::Armrest::TimeoutException)
```

(Note block argument to `expect`)

djberg96: Oops, thanks!

djberg96:
@bronaghs Ok, added some specs. :)

PR #221, opened 2016-10-05
Refactor the fetch_providers method

djberg96:
This deletes most of the code for the fetch_providers method. Instead, it now uses the ResourceProviderService#list method. Not only is the current code redundant, it doesn't cache the results. This causes a lot of unnecessary calls in the ManageIQ event runner, for example.

I opted to keep it a separate method for the sake of rspec testing. By keeping it separate, it's easier to stub responses for it.

The other change that was necessary was to default to nil in the provider_default_api_version method so that the set_service_api_version method can actually work properly.

Lastly I moved the cache_method require into the armrest.rb file where all the other libraries are required in case we want to use it within any other files.

### DIFF
--- a/lib/azure/armrest/configuration.rb
+++ b/lib/azure/armrest/configuration.rb
@@ -162,7 +162,7 @@ module Azure
 
       # Return the default api version for the given provider and service
       def provider_default_api_version(provider, service)
-        @provider_api_versions[provider.downcase][service.downcase]
+        @provider_api_versions[provider.downcase][service.downcase] rescue nil
       end
 
       # The name of the file or handle used to log http requests.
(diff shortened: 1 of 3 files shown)


djberg96:
@bzwei @bronaghs @blomquisg Please review.

lib/azure/armrest/configuration.rb
```diff
@@ -162,7 +162,7 @@ def token_expiration
 
       # Return the default api version for the given provider and service
       def provider_default_api_version(provider, service)
-        @provider_api_versions[provider.downcase][service.downcase]
+        @provider_api_versions[provider.downcase][service.downcase] rescue nil
       end
```
bronaghs: ```rescue nil``` is going to rescue from any exception. If this rescue statement is designed to handle the situation where the provider and service keys are not present then ```try``` method might be better option.

lib/azure/armrest/configuration.rb
```diff
@@ -266,22 +266,7 @@ def set_provider_api_versions
...
-        response = ArmrestService.send(
-          :rest_get,
-          :url         => uri.to_s,
-          :proxy       => proxy,
-          :ssl_version => ssl_version,
-          :ssl_verify  => ssl_verify,
-          :headers     => {
-            :content_type  => content_type,
-            :authorization => token
-          }
```
bronaghs: @djberg96 - Where are we setting these parameters now?

lib/azure/armrest/configuration.rb
djberg96: Ok, updated.

lib/azure/armrest/configuration.rb
bronaghs: @djberg96 - Do you mean the Configuration class constructor? Where is the ssl_verify set?

djberg96: @bronaghs Yes. The :ssl_verify accessor doesn't have an explicit default, so it defaults to nil. This is deliberate so that it defaults to whatever RestClient uses.

bronaghs:
@bzwei @blomquisg -  Anybody else want to chime in before I merge?

bzwei:
👍

PR #222, opened 2016-10-06
Added the get_template method

djberg96:
This adds the :get_template method to the TemplateDeploymentService class. This will allow us to get templates without resorting to inspecting deployment objects and then using raw URI calls.

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -179,6 +179,7 @@ module Azure
     # Initial class definitions. Reopen these classes as needed.
 
     class AvailabilitySet < BaseModel; end
+    class DeploymentTemplate < BaseModel; end
     class Event < BaseModel; end
     class ImageVersion < BaseModel; end
     class Offer < BaseModel; end
--- a/spec/template_deployment_service_spec.rb
+++ b/spec/template_deployment_service_spec.rb
@@ -42,6 +42,16 @@ describe "TemplateDeploymentService" do
       tds.create('deployname', 'groupname', {})
     end
 
+    it "defines a get_template method" do
+      expected = @req.merge(
+        :url     => url_prefix + "/deployname/exportTemplate?api-version=#{api_version}",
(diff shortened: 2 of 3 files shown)


djberg96:
@bzwei Need some help with testing please, my efforts were futile.

spec/template_deployment_service_spec.rb
```diff
@@ -42,6 +42,17 @@ def headers; {}; end
...
     end
 
+    it "defines a get_template method" do
+      allow(JSON).to receive(:parse).and_return({})
+      expected = @req.merge(
+        :url     => url_prefix + "/deployname/exportTemplate?api-version=#{api_version}",
+        :method  => :post,
+        :payload => ''
+      )
+      expect(RestClient::Request).to receive(:execute).with(expected).and_return('{}')
```
bzwei: if the test fails, try
`.and_return('{"template":{}}')`

lib/azure/armrest/model/base_model.rb
```diff
@@ -204,6 +204,7 @@ class TemplateDeployment < BaseModel
       attr_hash 'properties#parameters', 'properties#outputs'
     end
     class TemplateDeploymentOperation < TemplateDeployment; end
+    class TemplateDeploymentExport < TemplateDeployment; end
```
djberg96: Ok, done.

spec/template_deployment_service_spec.rb
djberg96: Well, seems to be passing as-is.

spec/template_deployment_service_spec.rb
```diff
@@ -42,6 +42,17 @@ def headers; {}; end
       tds.create('deployname', 'groupname', {})
     end
 
+    it "defines a get_template method" do
+      allow(JSON).to receive(:parse).and_return({})
```
djberg96: Ok, done.

bzwei:
👍 Thanks!

PR #224, opened 2016-10-13
Update the delete_blob method to return a ResponseHeaders object

djberg96:
This modifies the `StorageAccount#blob_delete` method to return a ResponseHeaders object. I also modified the `copy_blob` method to return the object created instead of just the headers (though the headers are included as part of the object).

There are a couple other minor tweaks, like using `concat` instead of "<<" + "flatten".

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -162,9 +162,7 @@ module Azure
           Container.new(Hash.from_xml(element.to_s)['Container'])
         end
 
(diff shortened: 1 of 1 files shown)


djberg96:
@bzwei If this looks alright, can you please merge?

bzwei:
Why `#concat`? There are too many ways to concat ruby array, which is hard to tell all the differences. I googled and it is said `<<` and `#concat` are aliased. But it seems you are saying `#concat` = `<<` + `#flatten`. Is it true?

bzwei:
👍

PR #225, opened 2016-10-14
Add get_by_id and get_subservice_by_id

djberg96:
This adds two methods to the ResourceGroupBasedService that will allow users to get objects based on an ID string without having to parse that information out themselves.

For example, a VM contains an ID for its NIC like this:

    "id": "/subscriptions/xyz/resourceGroups/some_group/providers/Microsoft.Network/networkInterfaces/some_nic"

With that information we can then get a model object like so:

    nic_id = vm.properties.network_profile.network_interfaces[0].id
    vms.get_by_id(nic_id)

This will also make deleting associated resources easier.

### DIFF
--- a/lib/azure/armrest/resource_group_based_service.rb
+++ b/lib/azure/armrest/resource_group_based_service.rb
@@ -2,6 +2,22 @@ module Azure
   module Armrest
     # Base class for services that need to run in a resource group
     class ResourceGroupBasedService < ArmrestService
+      # Used to map service name strings to internal classes
+      SERVICE_NAME_MAP = {
+        'availabilitysets'      => Azure::Armrest::AvailabilitySet,
+        'loadbalancers'         => Azure::Armrest::Network::LoadBalancer,
+        'networkinterfaces'     => Azure::Armrest::Network::NetworkInterface,
+        'networksecuritygroups' => Azure::Armrest::Network::NetworkSecurityGroup,
+        'publicipaddresses'     => Azure::Armrest::Network::IpAddress,
+        'storageaccounts'       => Azure::Armrest::StorageAccount,
+        'virtualnetworks'       => Azure::Armrest::Network::VirtualNetwork,
+        'subnets'               => Azure::Armrest::Network::Subnet,
+        'inboundnatrules'       => Azure::Armrest::Network::InboundNat,
+        'securityrules'         => Azure::Armrest::Network::NetworkSecurityRule,
+        'routes'                => Azure::Armrest::Network::Route,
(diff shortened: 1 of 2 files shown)


djberg96:

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -73,6 +73,85 @@ def list_all(filter = {})
...
+      #
+      # Example:
+      #
+      #   vms = Azure::Armrest::VirtualMachineService.new(conf)
+      #
+      #   vm = vms.get('your_vm', 'your_group')
+      #   nic_id = vm.properties.network_profile.network_interfaces[0].id
+      #   nic = vm.get_by_id(nic_id)
+      #
+      def get_by_id(id_string)
```
bzwei: suggest method name `get_associate_resource`

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -73,6 +73,85 @@ def list_all(filter = {})
...
+        info = parse_id_string(id_string)
+
+        api_version = configuration.provider_default_api_version(info[:provider], info[:service_name])
+        api_version ||= configuration.api_version
+
+        url = File.join(Azure::Armrest::RESOURCE, id_string) + "?api-version=#{api_version}"
+
+        model_class = case info[:service_name].downcase
```
bzwei: let's follow the indent style required by rubocop

djberg96: I think the indent style that rubocop wants is terrible.

lib/azure/armrest/resource_group_based_service.rb
djberg96: Hm, ok, but I've kept the originals as aliases because I like short method names. :)

djberg96:
@bzwei Ok, I modified the code so that it's a single method that works with either a service or subservice.

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -187,6 +238,37 @@ def list_in_all_groups
 
         array
       end
+
+      def get_model_class(service_name)
+        case service_name.downcase
```
bzwei: We can simply maintain a hash table for the lookup instead of long list of case/when

djberg96: @bzwei ok, done.

bzwei:
@djberg96 looks good. Can you address the warnings from rubocop?

djberg96:
@bzwei ok, fixed the rubocop warnings that I could.

bronaghs:
👍

PR #226, opened 2016-10-19
Clear all CacheMethod caches in the clear_caches method

djberg96:
This PR updates the `Configuration.clear_caches` method to clear the CacheMethod caches in addition to our own.

This is not only intuitive IMO, but also useful when regenerating VCR cassettes.


djberg96:
@bzwei Please review.

PR #227, opened 2016-11-02
Add method to delete associated resources

djberg96:
This PR adds the VirtualMachineService#delete_associated method. This will delete not only the VM, but resources associated with the VM as well. By default it will delete the NIC, the public IP, and the underlying image (vhd) file. You can optionally delete the network security group as well as the storage account that was backing it.

There's also a :verbose mode, in which case some extra logging is done. If there is no log, then STDOUT is used for the logger.

In cases where a user attempts to delete a resource that cannot be deleted (because something else is still attached to it), then it logs a warning and continues rather than failing. It is then up to the users to work out the issue, and delete the resource on their own.

I should add that we will need this for: https://bugzilla.redhat.com/show_bug.cgi?id=1348718

### DIFF
--- a/lib/azure/armrest/virtual_machine_service.rb
+++ b/lib/azure/armrest/virtual_machine_service.rb
@@ -121,6 +121,96 @@ module Azure
         vm_operate('powerOff', vmname, group)
       end
 
+      # Delete the VM and associated resources. By default, this will
+      # delete the VM, its NIC, the associated IP address, and the
+      # image file (.vhd) for the VM.
+      #
+      # If you want to delete any attached disks, the VM's underlying
+      # storage account, associated network security groups or
+      # availability set, you must explicitly specify them as an option.
+      #
+      # Because these resources could be associated with multiple VM's, you
+      # should be careful about deleting them. That said, attempting to
+      # delete these items with associated resources still attached will
+      # typically fail automatically. We make no guarantees here, though.
+      #
+      # Note that if all of your related resources are in a self-contained
+      # resource group, you do not necessarily need this method. You could
+      # just delete the resource group itself, which would automatically
+      # delete all of its resources
+      #
+      def delete_associated(vmname, vmgroup, options = {})
+        options = {
+          :network_interfaces     => true,
+          :ip_addresses           => true,
+          :os_disk                => true,
+          :data_disks             => false,
+          :network_security_group => false,
+          :attached_disks         => false,
+          :storage_account        => false,
+          :virtual_network        => false,
+          :scale_set              => false,
+          :verbose                => false
+        }.merge(options)
+
+        sas = Azure::Armrest::StorageAccountService.new(self.configuration)
+        nis = Azure::Armrest::Network::NetworkInterfaceService.new(self.configuration)
+        ips = Azure::Armrest::Network::IpAddressService.new(self.configuration)
+
+        vm   = get(vmname, vmgroup)
+        nics = vm.properties.network_profile.network_interfaces.map(&:id)
+
+        # Delete operations are asynchronous, so we have to poll after
+        # submitting delete operations.
+
+        delete(vmname, vmgroup)
+        puts "Deleting VM #{vmname}..." if options[:verbose]
+
+        sleep 10 while list(vmgroup).find{ |vm| vm.name == vmname }
+        puts "VM #{vmname} deleted" if options[:verbose]
+
+        nics.each do |nic_string|
+          nic_group = nic_string[/.*resourceGroups\/(.*?)\//i, 1]
+          nic_name = File.basename(nic_string)
+          nic = nis.get(nic_name, nic_group)
+
+          if options[:network_interfaces]
+            nis.delete(nic_name, nic_group)
+            puts "Deleting NIC #{nic_name}..." if options[:verbose]
+
+            sleep 10 while nis.list(nic_group).find{ |n| n.name == nic_name }
+            puts "NIC #{nic_name} deleted" if options[:verbose]
+
+            if options[:ip_addresses]
+              nic.properties.ip_configurations.each do |ip|
+                ip_string = ip.properties.public_ip_address.id
+                ip_group = ip_string[/.*resourceGroups\/(.*?)\//i, 1]
+                ip_name = File.basename(ip_string)
+                ips.delete(ip_name, ip_group)
+                puts "Deleting Public IP #{ip_name}..." if options[:verbose]
+              end
+            end
+          end
+        end
+
+        uri = Addressable::URI.parse(vm.properties.storage_profile.os_disk.vhd.uri)
+
+        storage_acct_name = uri.host.split('.').first
+        storage_acct      = sas.list_all.find{ |s| s.name == storage_acct_name }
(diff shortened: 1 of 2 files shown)


djberg96:
@bzwei @bronaghs Please review. I haven't added tests yet, so that's why it's marked WIP for now, but it's basically complete.

lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -121,12 +121,169 @@ def stop(vmname, group = configuration.resource_group)
...
+        uri = Addressable::URI.parse(vm.properties.storage_profile.os_disk.vhd.uri)
+
+        # The uri looks like https://foo123.blob.core.windows.net/vhds/something123.vhd
+        storage_acct_name = uri.host.split('.').first     # storage name, e.g. 'foo123'
+        storage_acct_disk = File.basename(uri.to_s)       # disk name, e.g. 'something123.vhd'
+        storage_acct_path = File.dirname(uri.path)[1..-1] # container, e.g. 'vhds'
+
+        # Must find it this way because the resource group information isn't provided
+        sas = Azure::Armrest::StorageAccountService.new(configuration)
+        storage_acct_obj  = sas.list_all.find { |s| s.name == storage_acct_name }
```
bronaghs: @djberg96 - this could be expensive

djberg96: I'll refactor this to look in the VM's resource group first, and only resort to the list_all if the first call doesn't work. The storage account will *usually* be in the same resource group.

Another option would be to cache those sas.list_all results.

lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -121,12 +121,169 @@ def stop(vmname, group = configuration.resource_group)
...
+              storage_acct_obj.delete_blob(blob.container, blob.name, key)
+            end
+
+            if extension == '.status' && blob.name.start_with?(vm.name)
+              if options[:verbose]
+                msg = "Deleting blob #{blob.container}/#{blob.name}"
+                Azure::Armrest::Configuration.log.info(msg)
+              end
+              storage_acct_obj.delete_blob(blob.container, blob.name, key)
+            end
```
bronaghs: @djberg96 - A bit of duplicated code here...

djberg96: Yep, good catch.

djberg96:
Ah, shoot, found a race condition that I need to deal with. There's a gap between when the VM is finally deleted and when the disk file is unlocked. If we try to delete it too soon, any attempt to delete it will fail.

djberg96:
@bronaghs Ok, refactored based on your comments.

Also, added a wait method for the underlying blob image to make sure that it's been released before we attempt to delete it.

lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -121,12 +121,187 @@ def stop(vmname, group = configuration.resource_group)
...
+
+      # Wait for blob to reach unlocked status.
+      def wait_for_blob(blob, max_time = 120)
+        cur_time = 0
+
+        while blob.properties.lease_status.downcase != 'unlocked'
+          cur_time += 10
+          break if cur_time > max_time
+          sleep cur_sleep
+        end
```
bronaghs: @djberg96 -  Where is ```cur_sleep``` set?

lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -121,12 +121,187 @@ def stop(vmname, group = configuration.resource_group)
...
+      end
+
+      # Wait for blob to reach unlocked status.
+      def wait_for_blob(blob, max_time = 120)
+        cur_time = 0
+
+        while blob.properties.lease_status.downcase != 'unlocked'
+          cur_time += 10
+          break if cur_time > max_time
```
djberg96: Yeah, my philosophy is that if it takes more than 2 minutes for the status to update, then just let it fail because it probably means something has gone wrong.

I could just do the "skip & log" thing if you think that's more appropriate.

djberg96: @bronaghs I've updated it to warn and skip in the latest incarnation.

djberg96:
@bzwei @bronaghs You can ignore the remaining rubocop warnings. The suppression of an exception is deliberate, and the warning about using "next" doesn't apply.

lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -121,12 +121,173 @@ def stop(vmname, group = configuration.resource_group)
...
+          keys = sas.list_account_keys(storage_account.name, storage_account.resource_group)
+          key  = keys['key1'] || keys['key2']
+          disk = sas.get_os_disk(vm)
+
+          # There's a short delay between deleting the VM and unlocking the underlying
+          # .vhd file by Azure. Therefore we sleep up to two minutes while checking.
+          if disk.x_ms_lease_status.casecmp('unlocked') != 0
+            sleep_time = 0
+
+            loop do
```
bzwei: Does `while sleep_time < 120 do` read better?

djberg96: Eh, I think it reads about the same, but I can change it.

bzwei:
@djberg96 I am not familiar with the logic how to delete resources. Your code in general looks good. My argument is - should we place the logic here or in miq? I have seen many APIs that come with wrappers for varies languages. They are all one-to-one translations. Do we consider our armrest gem such a wrapper? Why Azure API does not provide an API to delete resources associated with a VM? If Azure leaves the cleanup to the application, armrest gem may want to leave the handling to miq too. Just my two cents.

djberg96:
@bzwei Microsoft's philosophy is they're going to leave it up to you to decide what should be deleted because they have no way of knowing if you plan to re-use the IP address, storage account, network interface, etc, for future provisioning. I do think MS should at least provide some option you could attach to the request URL to delete the associated disk, but that doesn't exist yet.

Since I don't consider azure-armrest "just a REST wrapper", I like to provide functionality that I think users in the field will find useful. The question of how to delete associated resources (or, at least, the associated .vhd file) does get asked from time to time online, so I think it's worthwhile to add to the gem. I think we could potentially be saving a lot of users a lot of work.

djberg96:
@bzwei As for the logic, most of it revolves deleting resources "in order" and "when possible", e.g. you must delete a NIC before you can delete the IP, and you must also wait for the NIC to actually be deleted (since delete operations are asynchronous) before attempting to delete the IP. The blobs required a separate bit logic since they're files on a disk.

lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -121,12 +121,173 @@ def stop(vmname, group = configuration.resource_group)
...
+      #
+      # If the :verbose option is set to true, then additional messages are
+      # sent to your configuration log, or stdout if no log was specified.
+      #
+      # Note that if all of your related resources are in a self-contained
+      # resource group, you do not necessarily need this method. You could
+      # just delete the resource group itself, which would automatically
+      # delete all of its resources.
+      #
+      def delete_associated(vmname, vmgroup, options = {})
```
bzwei: how about `delete_associated_resources`?

PR #228, opened 2016-11-03
Add a log accessor and basic specs.

djberg96:
This adds a :log accessor to the ArmrestService class, the assumption being that it would be a Logger instance. This allows users to do logging for individual service classes without having to use the class level log, which is really only meant for logging the underlying http requests being made by rest-client.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -1,4 +1,5 @@
 require 'time'
+require 'logger'
 require_relative 'model/base_model'
 
 module Azure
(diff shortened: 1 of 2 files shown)


djberg96:

bzwei:
@djberg96 do we really need this? It is a gem library. The caller should log anything in the calling application. Besides `require 'logger'` seems strange since it is not referenced in your code at all.

djberg96:
@bzwei The idea was to use it internally for delete_associated in :verbose mode, since it felt strange to log it to the same logger as the http requests.

But, if you don't think it's useful, I'll go ahead and close it.

bzwei:
@djberg96 The existing logger for armrest gem logically is used internally by the gem to log anything it needs to, although it happens to log only http requests so far.
@bronaghs your thoughts?

PR #229, opened 2016-11-04
Updated Codeclimate gem

juliancheal:
Fixes this issue https://github.com/codeclimate/ruby-test-reporter#upgrading-from-pre-10-versions

- Added simplecov to spec helper
- Also added after script test to run codeclimate reporter

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,4 @@ matrix:
   - rvm: ruby-head
   - rvm: jruby-head
   fast_finish: true
+after_script: bundle exec codeclimate-test-reporter
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,8 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in azure-armrest.gemspec
 gemspec
+
+group :test do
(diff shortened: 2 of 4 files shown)


bronaghs:
@juliancheal - the documentation talks about setting CODECLIMATE_REPO_TOKEN
Should we do that somewhere?

Fryguy:
Since it's a hidden token, I set it in travis directly.

bronaghs:
OK, then this LGTM 
@Fryguy - looks good to you?

Fryguy:
@juliancheal I think we should also update the gemspec file with the version.. https://github.com/ManageIQ/azure-armrest/blob/825318b2f9d0420400d11ee3c7c2165c6349982c/azure-armrest.gemspec#L34

juliancheal:
@Fryguy ah thanks, didn't think to check the gemspec too.

juliancheal:
Is that really necessary though in the Gemspec as it's included in the Gemfile?

Fryguy:
yeah, technically not necessary, but good to be complete.

PR #230, opened 2016-11-07
Update regex for parse_id_string, add support for extensions.

djberg96:
This updates the parse_id_string method to handle hyphens and periods within resource groups and resource names, respectively. It also handles the "extension" service name, converting it to `VirtualMachineExtension`.

This is actually part 1 to fixing https://bugzilla.redhat.com/show_bug.cgi?id=1390715 because I need to get the extension information to get the correct storage account for metrics.


djberg96:
@bronaghs @bzwei Please review

bronaghs:
@djberg96 LGTM

PR #231, opened 2016-11-07
Modify the ResourceService list and list_all methods to get more that 1000 results

djberg96:
This modifies the ResourceService so that a user may optionally retrieve all resources for the `:list` and `:list_all` methods. Currently the results are capped at 1000.

Because this could retrieve a lot of data for a subscription, a user must specify `:all => true` for this behavior.

This PR required moving the :get_all_resources and :model_class methods from the ResourceGroupBasedService class to the ArmrestService class. That way all service classes inherit those methods.

### DIFF
--- a/lib/azure/armrest/resource_service.rb
+++ b/lib/azure/armrest/resource_service.rb
@@ -14,6 +14,9 @@ module Azure
       # resource group. You can optionally pass :top or :filter options as well
       # to restrict returned results.
       #
+      # Normally this is capped at 1000 results, but you may specify the :all
+      # option to get everything.
+      #
       # Examples:
       #
       #   rs = Azure::Armrest::ResourceService.new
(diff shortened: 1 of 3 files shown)


lib/azure/armrest/resource_service.rb
```diff
@@ -23,16 +26,29 @@ def initialize(configuration, options = {})
...
       end
 
       # Same as Azure::Armrest::ResourceService#list but returns all resources
       # for all resource groups.
       #
+      # Normally this is capped at 1000 results, but you may specify the :all
+      # option to get everything.
+      #
       def list_all(options = {})
         url = build_url(nil, options)
```
bzwei: will `options[:all]` have an impact on the resulting url?

lib/azure/armrest/resource_service.rb
```diff
@@ -23,16 +26,29 @@ def initialize(configuration, options = {})
...
         response = rest_get(url)
-        Azure::Armrest::ArmrestCollection.create_from_response(response, Azure::Armrest::Resource)
+
+        if options[:all]
+          get_all_results(response)
```
bzwei: what's the impact from existing applications - mainly manageiq? Does it need to add `options[:all]` after this update?

lib/azure/armrest/resource_service.rb
djberg96: No, it's using the returned nextlink property.

lib/azure/armrest/resource_service.rb
djberg96: Yes, it would need to be added within ManageIQ, though I don't know of anywhere that we use ResourceService directly right off. Do we?

bzwei: oh. This is for resource service only. Do you think `options[:all]` should be good for resource group based services too? If not, why by default one lists truly all, the other is truncated?

djberg96: Well, the other resource services only provide a subset of information, so I'm not generally worried about huge datasets being returned. Since this returns information for *all* resources, I felt it was warranted here.

PR #232, opened 2016-11-09
Add DiagnosticService class

djberg96:
Based originally on a request from Nick Catling, I've added a DiagnosticService class and some basic specs.

I've tested out a "live" example for both :get and :set and it works fine.


djberg96:

bzwei:
Looks good

PR #233, opened 2016-11-09
Add storage_account and virtual_disk methods to VirtualMachine model

djberg96:
This adds the storage_account and virtual_disk methods to the VirtualMachine model so that you can get that underlying information without having to search for it yourself. The storage_account method returns a StorageAccount object, while the virtual_disk method returns a BlobProperty object.

Example:

    vms = Azure::Armrest::VirtualMachineService.new(conf)
    sas = Azure::Armrest::StorageAccountService.new(conf)

    vm = vms.get(vmname, vmgroup)
    p vm.virtual_disk(sas)
    p vm.storage_account(sas)

This is ultimately part of an effort to simplify my solution for https://bugzilla.redhat.com/show_bug.cgi?id=1348718

Update: This is now being tackled from the StorageAccountService side:

    vm = vms.get('some_vm', 'some_rg')
    sas.get_from_vm(vm)
    sas.get_os_disk(vm)


djberg96:

bzwei:
Can we move the new methods to `VirtualMachineService` instead of requiring a vms passing into the `VirtualMachine` model?

djberg96:
@bzwei I suppose I can do that, though it's not as efficient in practice. What we really need is a nice way to access the service configuration settings from a model object.

lib/azure/armrest/storage_account_service.rb
```diff
@@ -203,6 +203,43 @@ def list_private_images(group = configuration.resource_group)
...
+
+        # The uri looks like https://foo123.blob.core.windows.net/vhds/something123.vhd
+        name = uri.host.split('.').first # storage name, e.g. 'foo123'
+
+        # Look for the storage account in the VM's resource group first. If
+        # it's not found, look through all the storage accounts.
+        begin
+          acct = get(name, vm.resource_group)
+        rescue Azure::Armrest::NotFoundException => err
+          acct = list_all.find { |s| s.name == name }
```
bzwei: This `list_all` requires only one API call, right? It takes a filter as the argument that you can use.

Do you see in real case that the account is in another resource group?

PR #234, opened 2016-11-09
Add a configuration singleton accessor 

djberg96:
This adds a :configuration accessor to the BaseModel class, and sets it within the various methods of the ResourceGroupBasedService. This will set the configuration information on a per model class basis, and would allow access to it within the models.

By allowing the models to access configuration information, models like StorageAccount can access and use that information directly, instead of having to set everything manually as we do now for StorageAccountService methods, for example.


djberg96:
After discussion with @bzwei, we're going to take a different approach.

PR #235, opened 2016-11-15
Add a service_name accessor

djberg96:
This adds a :service_name accessor. We store this information in an instance variable, e.g. "virtualMachines", but apparently never created an accessor for it.

This can be useful for URL building in internal methods.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -18,6 +18,9 @@ module Azure
       # Provider for service specific API calls
       attr_accessor :provider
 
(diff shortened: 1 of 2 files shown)


bzwei:
@djberg96 what is the use case? shouldn't the actual service class set the attribute? But I don't see it reflected in your code.

djberg96:
@bzwei The service class does set it. It's the 2nd argument to the constructor, so for VirtualMachineService it's "virtualMachines", for StorageAccountService it's "storageAccounts", etc.

bzwei:
I got it. It is straightforward. Thanks.

PR #236, opened 2016-11-17
Add separate SubscriptionService class and integrate it

djberg96:
This adds the Configuration#subscriptions method and removes redundant code from ArmrestService.

This is part of an effort towards implementing https://bugzilla.redhat.com/show_bug.cgi?id=1375285


bzwei:
@djberg96 Can we handle subscriptions the same way as providers? Maybe place the actual fetch logic in service class rather than in configuration?

djberg96:
@bzwei Let me think about it. We could do that, though it would require a separate constructor.

djberg96:
@bzwei Ok, I've added a SubscriptionService class and overhauled the Configuration class to use it. The big change is that subscription_id validation only happens if provided. Instead, the ArmrestService class verifies that the subscription_id is present.

The SubscriptionService defines its own constructor, ignoring that verification step. This means that, in practice, the subscription ID is *mostly* required, but that you can skip it for the SubscriptionService.

lib/azure/armrest/configuration.rb
```diff
@@ -198,36 +204,17 @@ def self.log=(output)
...
         end
 
-        if found.last.casecmp("enabled") != 0
-          warn "Subscription '#{found.first}' found but not enabled."
+        if found.state.casecmp('enabled') != 0
+          warn "Subscription '#{found.subscription_id}' found but not enabled."
         end
 
-        true
+        found.subscription_id
```
bzwei: No need to return `found.subscription_id` which is the same as existing `subscription_id`

lib/azure/armrest/configuration.rb
```diff
@@ -198,36 +204,17 @@ def self.log=(output)
...
-        found = subscriptions.find { |array| array.first == subscription_id }
+        found = @subscriptions.find { |sub| sub.subscription_id == subscription_id }
 
         unless found
           raise ArgumentError, "Subscription ID '#{subscription_id}' not found"
         end
 
-        if found.last.casecmp("enabled") != 0
-          warn "Subscription '#{found.first}' found but not enabled."
+        if found.state.casecmp('enabled') != 0
```
bzwei: good catch!

lib/azure/armrest/configuration.rb
```diff
@@ -108,23 +110,26 @@ def initialize(args)
...
         options.each { |key, value| send("#{key}=", value) }
 
-        unless client_id && client_key && tenant_id && subscription_id
-          raise ArgumentError, "client_id, client_key, tenant_id and subscription_id must all be specified"
+        unless client_id && client_key && tenant_id
+          raise ArgumentError, "client_id, client_key, and tenant_id must all be specified"
         end
 
         # Allows for URI objects or Strings.
         @proxy = @proxy.to_s if @proxy
```
bzwei: @djberg96 we need to re-visit all `attr_accessor` attributes as some are inappropriate. Do we allow them all to be modified after the configuration object is constructed? If not, we should make them read-only and allow them to be set only through constructor. For example, if subscription_id is set after initialization, then you won't get the chance to validate_subscription and fetch providers.

If we do allow them to be changed at anytime, I would change at least `@proxy` and `@subscription_id` to be `attr_reader` and provide customized setters, where you can force `@proxy` to be string and validate subscription and fetch providers.

djberg96: They are mostly set as accessors so that I can easily use hash args + send. I agree that this is probably not ideal but would like to keep that a separate PR.

bzwei: With a customized setter you can still use hash args + send, therefore I think it is appropriate to include the changes in this PR.

lib/azure/armrest/configuration.rb
bzwei: We don't need to purposely return anything because an error will be raised if the validation fails.

spec/spec_helper.rb
```diff
@@ -70,8 +70,13 @@ def setup_params
     Azure::Armrest::VirtualMachineSize.new(series2)
   ]
 
+  @subscriptions = [
+    Azure::Armrest::Subscription.new(:subscription_id => @sub, :state => 'Enabled')
+  ]
+
   allow_any_instance_of(Azure::Armrest::Configuration).to receive(:fetch_providers).and_return(@providers_response)
-  allow_any_instance_of(Azure::Armrest::Configuration).to receive(:validate_subscription).and_return(true)
+  allow_any_instance_of(Azure::Armrest::Configuration).to receive(:fetch_subscriptions).and_return(@subscriptions)
```
bzwei: Try to avoid `allow_any_instance_of`. Have you tried `allow(@conf)` instead?

lib/azure/armrest/configuration.rb
djberg96: Ok, I've added custom methods for `proxy=` and `subscription_id=`.

lib/azure/armrest/configuration.rb
djberg96: It no longer returns anything specifically.

lib/azure/armrest/configuration.rb
```diff
@@ -106,31 +105,47 @@ def initialize(args)
...
+        @tenant_id = options.delete(:tenant_id)
+        @client_id = options.delete(:client_id)
+        @client_key = options.delete(:client_key)
+
+        # Then set the remaining options automatically
         options.each { |key, value| send("#{key}=", value) }
 
-        unless client_id && client_key && tenant_id && subscription_id
-          raise ArgumentError, "client_id, client_key, tenant_id and subscription_id must all be specified"
+        unless client_id && client_key && tenant_id
```
bzwei: logically it makes more sense to move line 114 after the `unless` block

lib/azure/armrest/configuration.rb
```diff
@@ -106,31 +105,47 @@ def initialize(args)
...
-
         if user_token && user_token_expiration
           set_token(user_token, user_token_expiration)
         elsif user_token || user_token_expiration
           raise ArgumentError, "token and token_expiration must be both specified"
         end
 
-        @providers = fetch_providers
-        set_provider_api_versions
+        if subscription_id
```
bzwei: should move the `if` block to `def subscription_id=`

lib/azure/armrest/configuration.rb
djberg96: True.

lib/azure/armrest/configuration.rb
djberg96: Hm, I feel like that's overworking the subscription_id method. I don't think the provider information changes based on the subscription id, so it would potentially be making an unnecessary call.

bzwei: If I create the configuration without subscription_id and then set it in a following up call, the providers will never be filled. Therefore you should still move the code inside subscription_id setter. You can make it a separate method if you don't want to jam the setter method itself and check whether `@providers` blank before fetching.

djberg96: OK, I've moved that code into the subscription_id= method.

PR #238, opened 2016-11-30
Fix Billage::Usage model and update Usage#list method

djberg96:
This PR fixes the Billing::Usage model scoping (it was missing the Billing module), and updates the list method to return an ArmrestCollection instead of a plain array.

In addition, it now supports the `:all` option so that users can automatically retrieve more than 1k records. I also added a code example.


bzwei:
LGTM

PR #239, opened 2016-12-12
Add environment support

djberg96:
This PR allows US Government clients to acquire the correct auth token by adding an :environment accessor. Government clients use a different endpoint URL, so the proper resource URL is used depending on the environment.

The use of the term "environment" lines up with both Powershell and the Azure CLI.

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1403366

So in practice you could do:

    conf = Azure::Armrest::Configuration.new(
        :client_id   => 'xxx',
        :client_key  => 'yyy',
        :tenant_id   => 'zzz',
        :environment => 'US Government'
    )

    vms = Azure::Armrest::VirtualMachineService.new(conf)

    # You can also set it after the fact
    conf.environment = nil # would reset it to the public URL
    vms = Azure::Armrest::VirtualMachineService.new(conf)

### DIFF
--- a/lib/azure/armrest.rb
+++ b/lib/azure/armrest.rb
@@ -11,10 +11,13 @@ module Azure
   # The Armrest module mostly serves as a namespace, but also contains any
   # common constants shared by subclasses.
   module Armrest
-    # The default Azure resource
+    # The default (public) Azure resource
     RESOURCE = "https://management.azure.com/"
 
-    # The default authority resource
+    # The resource for US Government clients
+    GOV_RESOURCE = "https://management.usgovcloudapi.net/"
+
+    # The default (public) authority resource
     AUTHORITY = "https://login.windows.net/"
 
     # A common URI for all subclasses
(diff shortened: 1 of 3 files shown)


bzwei:
I prefer to pass in the url rather than hardcoded ones selected by `:environment`. This way if MS ever decide to add more urls, there is no need for the changes in armrest. The default can be always the public.

There should be reader only, no setter for the url, unless all other configuration attributes can remain the same except the url, which I doubt.

djberg96:
@bzwei Well, I already know there's another endpoint for China and the AD stuff, for example. But, a few things.

First is that in some cases, such as the AD stuff, the authority URL must be altered as well (confession: I'm not sure if that also needs to be modified here).  Second, I don't think an end user should have to know the endpoint URL's, and I think we should do our best to abstract that away, which means fewer mistakes for them. Last, I think if the user forgets the trailing "/" it will actually fail, so now you get into potentially having to mangle the URL.

Adding a few resource URL's with logic around them isn't very much work in practice.

bzwei:
@djberg96 OK. We can then use `:environment`, but why don't we simply predefine fixed choices rather than flexible `/US\s*?Government$/i`? For configurations we'd better to make most of the attributes read-only.

djberg96:
@bzwei Ok, I'll modify it.

bronaghs:
LGTM

PR #240, opened 2016-12-13
Add a common log method to ArmrestService

bzwei:
Ensure the RestClient.log to be an instance of Logger.

Move `VirtualMachineService#log_message` to base `ArmrestService` class. Slight modification - do not log anything unless the user provides a logger. Since this is a library, I feel we should not pollute console if the caller does not want to.

This common `log` method is private, meaning to be used internally by the service classes, not open to caller to write outside the gem.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -353,6 +353,10 @@ module Azure
       def model_class
         @model_class ||= Object.const_get(self.class.to_s.sub(/Service$/, ''))
       end
+
+      def log(level = "info", msg)
+        RestClient.log.try(level, msg)
+      end
     end # ArmrestService
   end # Armrest
 end # Azure
--- a/lib/azure/armrest/configuration.rb
+++ b/lib/azure/armrest/configuration.rb
@@ -215,6 +215,7 @@ module Azure
       # Sets the log to +output+, which can be a file or a handle.
       #
       def self.log=(output)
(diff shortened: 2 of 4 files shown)


lib/azure/armrest/configuration.rb
```diff
@@ -215,6 +215,7 @@ def self.log
       # Sets the log to +output+, which can be a file or a handle.
```
bzwei: @djberg96 shall we change the getter `def self.log` to return the logger only, instead of getting the log file name in the first place?

lib/azure/armrest/armrest_service.rb
```diff
@@ -353,6 +353,10 @@ def get_all_results(response)
       def model_class
         @model_class ||= Object.const_get(self.class.to_s.sub(/Service$/, ''))
       end
+
+      def log(level = "info", msg)
+        RestClient.log.try(level, msg)
```
djberg96: I thought it had to be a symbol. Can you confirm?

bzwei: Either string or symbol works

lib/azure/armrest/configuration.rb
djberg96: That's probably the way to go.

djberg96:
BTW, this won't break $azure_log will it?

bzwei:
The change will not break the use of `$azure_log` since it is an instance of `Logger`

djberg96:
Ok, LGTM.

PR #241, opened 2016-12-13
Add resource and authority accessors to Configuration class

djberg96:
This continues our goal of supporting US Gov for Azure by adding both `:resource` and `:authority` accessors to the Configuration class. If the `:environment` is set to 'usgov' then those accessors are set to the appropriate values automatically.

Service classes that were using the RESOURCE constants internally have been refactored to use the configuration accessor, and specs have been added.

Lastly, the resource for usgov needed a fix.

### DIFF
--- a/lib/azure/armrest/configuration.rb
+++ b/lib/azure/armrest/configuration.rb
@@ -67,6 +67,12 @@ module Azure
       # The environment in which to acquire your token.
       attr_reader :environment
 
+      # The authority URL used to acquire a valid token.
+      attr_accessor :resource
+
+      # The resource URL used to acquire a valid token.
+      attr_accessor :authority
+
       # Yields a new Azure::Armrest::Configuration objects. Note that you must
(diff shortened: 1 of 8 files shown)


lib/azure/armrest/configuration.rb
```diff
@@ -67,6 +67,12 @@ def self.cache_token(configuration)
       # The environment in which to acquire your token.
       attr_reader :environment
 
+      # The authority URL used to acquire a valid token.
+      attr_accessor :resource
```
bzwei: Is `resource` a too common word (just like service)? Maybe `resource_url`? Same for `authority` and constants if we want to make such renaming.

bzwei:
otherwise LGTM

lib/azure/armrest/configuration.rb
djberg96: I can rename those to :resource_url and :service_url, sure.

PR #242, opened 2016-12-14
Update environment= accessor to udpate auth and resource urls.

djberg96:
Minor update to last PR, this sets the authority_url and resource_url before calling fetch_token if the user calls environment= explicitly.

### DIFF
--- a/lib/azure/armrest/configuration.rb
+++ b/lib/azure/armrest/configuration.rb
@@ -124,7 +124,7 @@ module Azure
         # Delay this to avoid a double call
         @environment = options.delete(:environment)
 
-        if @environment.to_s.casecmp(Azure::Armrest::USGOV_ENVIRONMENT) == 0
+        if @environment.to_s.downcase == Azure::Armrest::USGOV_ENVIRONMENT
           options[:authority_url] = Azure::Armrest::USGOV_AUTHORITY
           options[:resource_url]  = Azure::Armrest::USGOV_RESOURCE
         end
@@ -202,16 +202,9 @@ module Azure
       # argument to 'USGov'. Otherwise, set it to nil.
       #
       def environment=(env)
-        if env != environment
-          if env.to_s.casecmp(Azure::Armrest::USGOV_ENVIRONMENT) == 0
-            authority_url = Azure::Armrest::USGOV_AUTHORITY
-            resource_url = Azure::Armrest::USGOV_RESOURCE
-          else
-            authority_url = Azure::Armrest::AUTHORITY
-            resource_url = Azure::Armrest::RESOURCE
-          end
-          fetch_token
-        end
+        return if env == environment
+        set_auth_and_resource_urls(env)
+        fetch_token
         @environment = env
       end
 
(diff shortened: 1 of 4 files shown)


lib/azure/armrest/configuration.rb
```diff
@@ -202,7 +202,16 @@ def token_expiration
       # argument to 'USGov'. Otherwise, set it to nil.
       #
       def environment=(env)
-        fetch_token if env != environment
+        if env != environment
```
bzwei: `return if env == environment` then the rest of the code have fewer indent.

lib/azure/armrest/configuration.rb
```diff
@@ -202,7 +202,16 @@ def token_expiration
...
+          if env.to_s.casecmp(Azure::Armrest::USGOV_ENVIRONMENT) == 0
```
bzwei: Better use `case when` statement so in the future we can easily add other environments without lots of if/else

djberg96: Yeah, I was thinking I should split out the logic to another method anyway.

bzwei:
@djberg96 please refactor the constructor to reuse code by calling `environment=`. 
What's the environment if user provides resource_url and authorize_url?

djberg96:
@bzwei Originally I was going to do that, but it causes a double fetch.

lib/azure/armrest/configuration.rb
djberg96: True.

lib/azure/armrest/configuration.rb
```diff
@@ -124,7 +124,7 @@ def initialize(args)
         # Delay this to avoid a double call
         @environment = options.delete(:environment)
 
-        if @environment.to_s.casecmp(Azure::Armrest::USGOV_ENVIRONMENT) == 0
+        if @environment.to_s.downcase == Azure::Armrest::USGOV_ENVIRONMENT
```
bzwei: We don't need any special handling for `@environment`, just let it naturally fall through the setter for `environment`. So you can delete line 109, 110, 125, and block 127.

lib/azure/armrest/configuration.rb
```diff
@@ -202,7 +202,9 @@ def token_expiration
       # argument to 'USGov'. Otherwise, set it to nil.
       #
       def environment=(env)
-        fetch_token if env != environment
+        return if env == environment
+        set_auth_and_resource_urls(env)
+        fetch_token
```
bzwei: We should make `environment=` a private method. Then there is no need to `fetch_token` here since `environment=` method is  is called only once. We can also merge `set_auto_and_resource_urls` back to this method since there is no other logic

lib/azure/armrest/configuration.rb
bzwei: Then we can still delete line 109 and 110.

bzwei: Forgive me for not thinking thoroughly. The above suggestion is still buggy. We need to think the following combination:

1. no urls, no environment (the most common one)
2. no urls, but with environment (such as usgov)
3. with urls, no environment (user case?)
4. with urls and environment (do we allow because they may conflict?)

lib/azure/armrest/configuration.rb
```diff
@@ -248,11 +235,11 @@ def subscriptions
...
         case env.to_s.downcase
           when Azure::Armrest::USGOV_ENVIRONMENT
-            authority_url = Azure::Armrest::USGOV_AUTHORITY
-            resource_url = Azure::Armrest::USGOV_RESOURCE
+            @authority_url = Azure::Armrest::USGOV_AUTHORITY
+            @resource_url = Azure::Armrest::USGOV_RESOURCE
           else
-            authority_url = Azure::Armrest::AUTHORITY
-            resource_url = Azure::Armrest::RESOURCE
+            @authority_url = Azure::Armrest::AUTHORITY
```
bzwei: no `else` clause is needed since default is already set in constructor

spec/configuration_spec.rb
```diff
@@ -53,13 +53,6 @@
       options[:token_expiration] = Time.now.utc + 1.month
       expect(described_class.new(options).token).to eq('token_string')
     end
-
-    it 'sets the authority url and resource url to appropriate values for US Gov' do
```
bzwei: why did you remove this valid spec test?

djberg96: Ok, restored.

PR #243, opened 2016-12-14
ArmrestService#wait accepts max_time = 0 to disable timeout

bzwei:
With this refactoring the `wait` can wait without timeout limitation until the status actually turns into succeeded.


djberg96:
👍

PR #244, opened 2016-12-14
Allow to delete all resources generated by a deployment

bzwei:
This PR implements to delete all resources created by the deployment together with the deployment itself.

Because the resources are fetched with IDs, so we can directly delete them using the IDs without go through each individual service to make deletion.

Since it is not easy to resolve the dependencies among resources, some of them cannot be deleted in the first run because they may be used by others. We will retry a number of times, with the worst case each run delete only one resource.

The `#delete_associated_resources` method is a synchronous call. It will wait until all resources are deleted. If in the end there are resources left undeleted, they are logged with message type `error`.


lib/azure/armrest/template_deployment_service.rb
```diff
@@ -49,6 +49,68 @@ def get_template(deploy_name, resource_group = configuration.resource_group)
...
       end
+
+      # Delete a deployment and all associated resources that were generated by the
+      # deployment.
+      #
+      # This is a synchronous call. It waits until all deletions complete
+      def delete_associated_resources(deploy_name, resource_group = configuration.resource_group)
+        operations = list_deployment_operations(deploy_name, resource_group)
+        resource_urls = operations.collect do |op|
+          if op.properties.provisioning_operation == 'Create'
```
djberg96: I'm super paranoid about Azure changing the case, as I have seen both upper and lower case responses, depending on the situation. I would recommend using casecmp('Create') instead.

lib/azure/armrest/template_deployment_service.rb
```diff
@@ -49,6 +49,68 @@ def get_template(deploy_name, resource_group = configuration.resource_group)
...
+            resource_id = op.properties.target_resource.id
+            "#{Azure::Armrest::RESOURCE}#{resource_id}?api-version=#{api_version}"
```
djberg96: You don't want to use the RESOURCE constant, since it would fail for USGov, for example. Use configuration.resource_url.

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -157,40 +160,51 @@ def delete(name, rgroup = configuration.resource_group)
...
-          providers/
-          (?<provider>[^\/]+)?/
-          (?<service_name>[^\/]+)?/
-          (?<resource_name>[^\/]+)
-          (/(?<subservice_name>[^\/]+)?/(?<subservice_resource_name>[^\/]+))*
+          subscriptions/(?<subscription_id>[^\/]+)?/
+          (resourceGroups/(?<resource_group>[^\/]+)?/)?
+          (providers/(?<provider>[^\/]+)?/)?
+          ((?<service_name>[^\/]+)?/(?<resource_name>[^\/]+)/)?
+          ((?<subservice_name>[^\/]+)?/(?<subservice_resource_name>[^\/]+))?
```
bzwei: Every parameter is now optional except subscription

djberg96:
👍

PR #246, opened 2016-12-15
Fix poll and wait methods

bzwei:
Last refactoring #243 introduced a bug, where `while` loop should be `until` loop.

There is a case for `poll` method that needs special attention. The poll may return an empty body that cannot be resolved into a status.

This PR addressed both issues.


djberg96:
👍

PR #248, opened 2016-12-16
Fix regex for #parse_id_string

bzwei:
Switch ending slash to leading slash. Fix the bug that cannot parse id without a subservice.


djberg96:
👍

PR #249, opened 2016-12-20
Add Environment class and predefined Public and USGovernment instances

djberg96:
This PR adds an Environment class that contains all the relevant URL information. It is based mostly on the Azure endpoint mapping docs*, as well as some inspiration from one or two other Azure libraries out in the wild.

The Configuration#environment accessor now takes an instance of an Environment class instead of a string. Two pre-defined instances have been declared as well - Environment::Public and Environment::USGovernment. We can add more as needed later.

Most of the constants have been removed as they are no longer needed.

https://docs.microsoft.com/en-us/azure/azure-government-developer-guide#a-nameendpointaendpoint-mapping

### DIFF
--- a/lib/azure/armrest/configuration.rb
+++ b/lib/azure/armrest/configuration.rb
@@ -64,14 +64,9 @@ module Azure
       # Maximum number of threads to use within methods that use Parallel for thread pooling.
       attr_accessor :max_threads
 
-      # The environment in which to acquire your token.
-      attr_reader :environment
-
-      # The authority URL used to acquire a valid token.
-      attr_accessor :resource_url
-
-      # The resource URL used to acquire a valid token.
-      attr_accessor :authority_url
+      # The environment object which determines various endpoint URL's. The
+      # default is Azure::Armrest::Environment::Public.
+      attr_accessor :environment
 
       # Yields a new Azure::Armrest::Configuration objects. Note that you must
       # specify a client_id, client_key, tenant_id. The subscription_id is optional
@@ -106,8 +101,7 @@ module Azure
           :proxy         => ENV['http_proxy'],
           :ssl_version   => 'TLSv1',
           :max_threads   => 10,
-          :authority_url => Azure::Armrest::AUTHORITY,
-          :resource_url  => Azure::Armrest::RESOURCE
+          :environment   => Azure::Armrest::Environment::Public
         }.merge(args.symbolize_keys)
 
         # Avoid thread safety issues for VCR testing.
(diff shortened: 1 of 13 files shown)


lib/azure/armrest/configuration.rb
```diff
@@ -64,14 +64,9 @@ def self.cache_token(configuration)
...
-      attr_accessor :resource_url
-
-      # The resource URL used to acquire a valid token.
-      attr_accessor :authority_url
+      # The environment object which determines various endpoint URL's. The
+      # default is Azure::Armrest::Environment::Public.
+      attr_accessor :environment
 
       # Yields a new Azure::Armrest::Configuration objects. Note that you must
       # specify a client_id, client_key, tenant_id. The subscription_id is optional
```
bronaghs: @djberg96 - is the subscription_id optional in the gem?

lib/azure/armrest/environment.rb
```diff
@@ -0,0 +1,124 @@
...
+        :publish_settings_file_url,
+        :resource_manager_url,
+        :service_management_url,
+        :sql_database_dns_suffix,
+        :storage_suffix,
+        :traffic_manager_dns_suffix
+      ]
+
+      # The Environment name
+      attr_reader :name
```
bzwei: How about `VALID_KEYS.each { |k| attr_reader k }` to avoid repeatedly typing these keys?

djberg96: Because I wouldn't be able to document them.

bzwei: you can append the comment to each line in the hash

lib/azure/armrest/configuration.rb
djberg96: @bronaghs Not sure what you mean. It's *technically* optional in the Configuration constructor, but almost all Service classes require it. It's not required for the SubscriptionService class, though.

lib/azure/armrest/environment.rb
djberg96: I don't think that will do anything for Doc generators, though.

lib/azure/armrest/configuration.rb
```diff
@@ -307,7 +279,7 @@ def fetch_providers
       end
 
       def fetch_token
-        token_url = File.join(authority_url, tenant_id, 'oauth2/token')
+        token_url = File.join(environment.authority_url, tenant_id, 'oauth2', 'token')
```
bronaghs: @djberg96 - i was referring to the change to ``` 'oauth2/token'```

djberg96: @bronaghs Oh, haha, sorry. That's a harmless change, I just thought it looked better.

lib/azure/armrest/configuration.rb
bronaghs: @djberg96 - I think you understood me correctly, I was surprised to learn the subscription id is optional.

lib/azure/armrest/configuration.rb
bronaghs: @djberg96  - I think so too :-)

lib/azure/armrest/configuration.rb
djberg96: @bronaghs All that means is that you can call `Configuration.new` without one, but if you try to pass a `conf` object to, say, a VirtualMachineService and it isn't defined, it will fail.

In effect, this means that the subscription_id requirement is *delayed* rather than optional in most cases. Why did we do this? Because technically you don't need it to get the initial token. This was desired because we wanted to be able to get a list of subscriptions without first having to provide one.

And we wanted *that* ultimately to facilitate cloud discovery in MIQ.

bronaghs:
Im good with this - nice Dan! 👍

lib/azure/armrest/configuration.rb
bronaghs: @djberg96 -  Got it - thanks!

PR #251, opened 2016-12-21
Explicit return if subscription ID is nil or blank

djberg96:
For the `Configuration#subscription_id=` method, do not attempt validation or gather provider info if the argument is nil or blank since it will fail.


bronaghs:
@djberg96 - 
What about the SubscriptionService class? The subscription ID will be null

djberg96:
@bronaghs The SubscriptionService class is the one service class that doesn't require it. This allows us to get a list of available subscriptions without first requiring a subscription.

PR #252, opened 2016-12-22
Add environment name to hash method

djberg96:
This updates the `hash` method to include the environment name for cases where users are in mixed environments. Otherwise conflicts can arise from cached tokens that were generated for a  different environment.


bzwei:
👍

PR #253, opened 2017-01-05
Update MetricsService class to include newer REST API

djberg96:
This updates the MetricsService class so that it supports the newer API for metrics and metrics definitions.

The current `list` method, probably somewhat poorly chosen in hindsight, will remain for backwards compatibility. The `list_metrics` and `list_definitions` have been added to support the newer API. I kept the original method because it returns different output. The hardcoded api-version string is now only used for the older method.

In addition, I converted the `list` method to return an ArmrestCollection.

Originally written to try to work get metrics working for usgov regions (it still doesn't), this is a much easier way to get metrics information for public environments than our current approach, which requires accessing the storage account directly.

### DIFF
--- a/lib/azure/armrest/insights/metrics_service.rb
+++ b/lib/azure/armrest/insights/metrics_service.rb
@@ -5,12 +5,14 @@ module Azure
         # Creates and returns a new MetricsService object.
         #
         def initialize(armrest_configuration, options = {})
-          options['api_version'] = '2014-04-01' # Must hard code for now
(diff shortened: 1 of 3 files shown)


djberg96:

djberg96:
@bzwei I added a couple basic tests so this one is ready to go.

bzwei:
@djberg96 does it make sense to split it into two services, one for metrics service, the other definition service?

djberg96:
@bzwei Yes, but that will cause breakage.

PR #254, opened 2017-01-12
[WIP] Add support for ActiveDirectory services.

djberg96:
This PR adds support for ActiveDirectory services, starting with UserService and GroupService classes. These are descended from a base ActiveDirectoryService class which contains a set of common methods that subclasses can use, similar to the ResourceGroupBasedService.

One significant change is the addition of the ActiveDirectoryConfiguration class. Because the endpoint for acquiring an authentication token is different than the regular configuration class, some changes were necessary to accommodate that, notably modifying the fetch_token method and adding the get_token method.

This is not the only possible approach, but the alternatives require making multiple calls for each endpoint with updates to our hashing and caching approach.

Anyway, as things stand now you can do something like this:

```
ad_conf = Azure::Armrest::ActiveDirectoryConfiguration.new(cred)
ads = Azure::Armrest::ActiveDirectory::UserService.new(ad_conf)
gds = Azure::Armrest::ActiveDirectory::GroupService.new(ad_conf)

p ads.list_all
p gds.list_all
```


lib/azure/armrest/active_directory/configuration.rb
```diff
@@ -0,0 +1,47 @@
...
+        # Fetch and set the token and token expiration before calling super
+        # so that it doesn't try to fetch it again from the wrong resource.
+
+        unless options[:token] || options[:token_expiration]
+          fetch_token
+          options[:token] = @token
+          options[:token_expiration] = @token_expiration
+        end
+
+        super(options)
```
bzwei: Some of above code are already in super. What you need is to prepare `options` to be used by `super`, but avoid setting values that will be repeatedly set by `super`.

djberg96: The problem I was having is that some of those values need to be set before the call to `fetch_token`, and thus before the call to super.

PR #255, opened 2017-01-17
Refactor RoleAssignmentService and RoleDefinitionService

djberg96:
The current RoleAssignmentService and RoleDefinitionService currently don't work, except for the list method. This is mainly due to a change in the Azure API and updates to the documentation that explain that the requests for this type of information should be scoped to the subscription, resource or resource group level.

Consequently, an abstract base class was created that both service classes now subclass from, and new methods were added to indicate at which scope each method is being made.


djberg96:
@bzwei Please review.

PR #258, opened 2017-02-23
Replace cache_method with memoist gem

Fryguy:
@djberg96 Please review.  I've chosen memoist because it solves the same problem of "caching methods that have parameters", but it just uses "normal" memory memoization via ivars under the covers.  Additionally, it's already a dependency of manageiq (we use it for caching of Relationship records).

Note that this removes the `cache_time` class accessor method.  I don't see it used anywhere in manageiq nor in manageiq-providers-amazon, which means it must always be defaulting to 24 hours.  At 24 hours for a instance method cache, that's practically infinity, so I'm not sure what purpose it actually serves.  Additionally, it was implemented as a class-level shared timeout, even though setting it is at an instance level, which is strange.

The first commit was just so I wouldn't get "double gem" warnings when bundling, so technically it can be a separate PR if you want.

### DIFF
--- a/lib/azure/armrest/resource_provider_service.rb
+++ b/lib/azure/armrest/resource_provider_service.rb
@@ -1,89 +1,54 @@
 module Azure
   module Armrest
     class ResourceProviderService < ArmrestService
+      extend Memoist
+
       # The provider used in http requests. The default is 'Microsoft.Resources'
       attr_reader :provider
 
-      # The amount of time in seconds to cache certain methods. The default is 24 hours.
-      @cache_time = 24 * 60 * 60
-
-      class << self
-        # Get or set the cache time for all methods. The default is 24 hours.
-        attr_accessor :cache_time
-      end
-
       # Creates and returns a new ResourceProviderService object.
       #
-      # Note that many ResourceProviderService instance methods are cached. You
-      # can set the cache_time for certain methods in the constructor, but keep in
-      # mind that it is a global setting for the class. You can also set this
-      # at the class level if desired. The default cache time is 24 hours.
+      # Note that many ResourceProviderService instance methods are cached.
       #
       # You can also set the provider. The default is 'Microsoft.Resources'.
       #
       def initialize(configuration, options = {})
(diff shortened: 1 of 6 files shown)


djberg96:
@bzwei How does it look?

djberg96:
@Fryguy We were explicitly setting it to 24 hours for one resource class' methods in particular. Is there a way to specify the time at which the cache will expire with memoist. While new resource providers are rare, they do get added once in a great while. I think I was planning on doing something similar with flavors.

lib/azure/armrest/resource_provider_service.rb
```diff
@@ -1,89 +1,54 @@
...
-
-      private :_list
-
-      # Cannot directly cache list method, because Marshal used by cache_method
-      # cannot dump anonymous classes, including the ones derived from
-      # Azure::Armrest::BaseModel.
-      #
-      # The same applies to other cached methods in this class.
-      cache_method(:_list, cache_time)
+      memoize :list
```
bzwei: Has this been actually tested? `cache_method` cannot cache Armrest model instances, and it is the reason we have to cache the underline hash object. Please make sure `memoize` does not have the same limitation.

Fryguy:
@djberg96 But was is the purpose at an instance level?  Are you keeping an instance of ResourceProviderService around for more than 24 hours?  If so, why?

lib/azure/armrest/resource_provider_service.rb
Fryguy: > cache_method cannot cache Armrest model instances, and it is the reason we have to cache the underline hash object.

That's because cache-method use `marshal_dump` which can't dump singletons.  Memoist just uses a normal hash cache.

Fryguy:
> Is there a way to specify the time at which the cache will expire with memoist.

There is not, which is why I was hoping to get rid of it.  Memoist does have a `flush_cache` method that can be called manually.  If we need the timeout, then this probably won't work, but really like to understand why we need it at all because I can't see why it's necessary.

djberg96:
@bzwei I did some general live futzing, and it appears to work fine.

PR #260, opened 2017-03-17
Add a DiskService

djberg96:
This adds a DiskService so that we can interface with managed disks. Azure is moving away from individual .vhd's on individual storage accounts to managed disks.

https://docs.microsoft.com/en-us/azure/storage/storage-managed-disks-overview

This is ultimately part of an effort to solve this BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1431361


djberg96:

Expect classes for Images and Snapshots soon as well. :)

PR #261, opened 2017-03-17
Add ImageService

djberg96:
This adds the ImageService class. This is another element of the managed accounts that Azure is moving towards.


djberg96:

lib/azure/armrest.rb
```diff
@@ -26,6 +26,7 @@ module Armrest
 require 'azure/armrest/storage_account_service'
 require 'azure/armrest/availability_set_service'
 require 'azure/armrest/disk_service'
+require 'azure/armrest/image_service'
```
bzwei: What's the difference of image_service from virtual_machine_image_service?

djberg96: @bzwei The VirtualMachineImageService is not a resource group based service. It is designed to get publisher/offer/sku information. The ImageService (or, soon to be Storage::ImageService) is a resource group based service. It provides information for images that are currently used by VM's using managed disks.

PR #262, opened 2017-03-17
Add SnapshotService

djberg96:
This adds a SnapshotService class. This is to keep up with Azure's move from individual storage accounts to managed disks.


djberg96:

bzwei:
@djberg96 Have you thought about adding a storage lay like `Azure::Armrest::Storage::AbcService`? Should we group all new services due to Azure's movement to managed disks to this folder?

djberg96:
@bzwei I thought about putting it under a "ManagedDisk" namespace, since that's currently how the Azure docs organize it. I could see using "Storage" as well to be more generic, since MS could change the name.

djberg96:
@bzwei Ok, I moved both Snapshot and Disk under the Storage namespace. I'll move Image after this is merged.

bzwei:
LGTM. I guess the existing `StorageService` has not been widely used 😄

PR #266, opened 2017-03-23
Change get_template to return a string

djberg96:
The `TemplateDeploymentService#get_template` method currently tries to generate and return a `DeploymentTemplate` model object. However, this is problematic because a template can contain arbitrary text that our base model cannot translate into a real Ruby constant.

Consequently, this method has been altered to just return a plain JSON string instead of a model object. While this technically breaks backwards compatibility, the reality is that this method is likely unused in the wild since it's just a matter of time before it fails, and that the raw JSON is what people are after anyway. ~~~That said, I have left the `DeploymentTemplate` model in the code for those people who may want to try and convert it to an object on their own.~~~

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1434988


lib/azure/armrest/template_deployment_service.rb
```diff
@@ -38,16 +38,13 @@ def get_deployment_operation(op_id, deploy_name, resource_group = configuration.
...
-      # If you want the plain JSON text then call .to_json on the returned object.
+      # Returns the raw json template for the given deployment as a string.
       #
       def get_template(deploy_name, resource_group = configuration.resource_group)
         validate_resource_group(resource_group)
         validate_resource(deploy_name)
         url = build_url(resource_group, deploy_name, 'exportTemplate')
-        response = JSON.parse(rest_post(url))['template']
-        DeploymentTemplate.new(response)
+        JSON.parse(rest_post(url))['template'].to_json
```
bzwei: We should remove `DeploymentTemplate` from base_model.rb.

djberg96: @bzwei I thought about it, but thought it might be nice for people who do want to convert it. Or do you think that will just lead to confusion?

djberg96:
@bzwei Ok, removed that model.

Issue #268, opened 2017-04-07
Add ability to get raw blob information on managed disks


djberg96:
We'll need to modify the `Storage::DiskService` class in order to support SSA for ManageIQ for VM's that use managed disks. The Azure REST API provides separate a separate method for getting raw blob information on managed disks:

https://docs.microsoft.com/en-us/rest/api/manageddisks/disks/disks-grant-access

It's a 3 step process. First, you have to get the location URL from the response header, then you have to get the URL that includes the SAS token, and finally you make a request with that URL to get the blob information.

PR #269, opened 2017-04-07
Add get_blob_raw method.

djberg96:
This allows us to get the raw blob data for a VM that's backed by a managed disk instead of a storage account. I also added a ResponseBody model for convenience.

This is similar to the StorageAccount model method  of the same name but there are some internal differences. Mainly, this method requires multiple calls internally to work.

Addresses https://github.com/ManageIQ/azure-armrest/issues/268

~~~WIP for now until I can get some specs made. I may also want to call `EndGetAccess` explicitly, but I'm not sure yet.~~~

### DIFF
--- a/lib/azure/armrest/storage/disk_service.rb
+++ b/lib/azure/armrest/storage/disk_service.rb
@@ -11,6 +11,114 @@ module Azure
         def initialize(configuration, options = {})
           super(configuration, 'disks', 'Microsoft.Compute', options)
         end
+
+        # Get the raw blob information for a managed disk. This is similar to
+        # the StorageAccount#get_blob_raw method, but applies only to a managed
+        # disk, whereas that method applies only to an individual storage
+        # account.
+        #
+        # As with the Storage#get_blob_raw method, you should pass a :range,
+        # :start_byte, :end_byte or :length option. If you want the entire
+        # image you must pass the :entire_image option, though this is generally
+        # not recommended. Unlike the Storage#get_blob_raw method, this method
+        # does not support the :date parameter.
+        #
+        # The +options+ are as follows:
+        #
+        #   :range        => A range of bytes you want, e.g. 0..1023 to get first 1k bytes
+        #   :start_byte   => The starting byte number that you want to collect bytes for. Use
+        #                    this in conjunction with :length or :end_byte.
+        #   :end_byte     => The ending byte that you want to collect bytes for. Use this
+        #                    in conjunction with :start_byte.
+        #   :length       => If given a :start_byte, specifies the number of bytes from the
+        #                    the :start_byte that you wish to collect.
+        #   :entire_image => If set, returns the entire image in bytes. This will be a long
+        #                    running request that returns a large number of bytes.
+        #
+        # You may also pass a :duration parameter, which indicates how long, in
+        # seconds, that the privately generated SAS token should last. This token
+        # is used internally by requests that are used to access the requested
+        # information. By default it lasts for 1 hour.
+        #
+        # Get the information you need using:
+        #
+        # * response.body    - blob data (the raw bytes).
+        # * response.headers - blob metadata (a hash).
+        #
+        # Example:
+        #
+        #   vms = Azure::Armrest::VirtualMachineService.new(conf)
+        #   sds = Azure::Armrest::Storage::DiskService.new(conf)
+        #
+        #   vm = vms.get(vm_name, vm_resource_group)
+        #   os_disk = vm.properties.storage_profile.os_disk
+        #
+        #   disk_id = os_disk.managed_disk.id
+        #   disk = sds.get_by_id(disk_id)
+        #
+        #   # Get the first 1024 bytes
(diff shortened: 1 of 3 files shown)


lib/azure/armrest/storage/disk_service.rb
```diff
@@ -11,6 +11,115 @@ class DiskService < ResourceGroupBasedService
         def initialize(configuration, options = {})
           super(configuration, 'disks', 'Microsoft.Compute', options)
         end
+
+        # Get the raw blob information for a managed disk. This is similar to
```
bzwei: @djberg96 I always like your documentation with examples. Very informative and practical. 👏

djberg96: Thanks. :)

lib/azure/armrest/storage/disk_service.rb
```diff
@@ -11,6 +11,115 @@ class DiskService < ResourceGroupBasedService
...
+        #
+        # You may also pass a :durationInSeconds parameter, which indicates
+        # how long the privately generated SAS token should last. This token
+        # is used internally by requests that are used to access the requested
+        # information. By default it lasts for 1 hour.
+        #
+        # Get the information you need using:
+        #
+        # * response.body    - blob data.
+        # * response.headers - blob metadata.
```
bzwei: What is the type of `response.headers` - is it `ResponseHeaders`? 
Same question for `response.body`.

lib/azure/armrest/storage/disk_service.rb
```diff
@@ -11,6 +11,115 @@ class DiskService < ResourceGroupBasedService
...
+          unless sas_url
+            msg = "Unable to find an SAS URL for #{disk_name}/#{resource_group}"
+            raise Azure::Armrest::NotFoundException.new(response.code, msg, response.body)
+          end
+
+          # The same restrictions that apply to the StorageAccont method also apply here.
+          range_str = nil
+
+          if range
+            range_str = "bytes=#{range.min}-#{range.max}"
```
bzwei: You can have logic to create a `range` object based on all combinations for start, end, length etc, and convert to `range_str` with one shot.

lib/azure/armrest/storage/disk_service.rb
```diff
@@ -11,6 +11,115 @@ class DiskService < ResourceGroupBasedService
...
+            if end_byte
+              range_str << end_byte.to_s
+            elsif length
+              range_str << (start_byte + length - 1).to_s
+            end
+          end
+
+          headers = {}
+
+          if range_str
```
bzwei: Does it looks slightly better this way (for Ln 104-110)?
```
raise ArgumentError, "..." unless entire_image || range_str
headers = {'x-ms-range' => range_str}.compact
```

djberg96:
@bzwei Ok, I made the suggested changes. I'm not a huge fan of multiple hash arguments, but odds are that most people will use the defaults for the timeout, so I can live with it here.

lib/azure/armrest/storage/disk_service.rb
```diff
@@ -11,6 +11,112 @@ class DiskService < ResourceGroupBasedService
...
+          sas_url = body.try(:properties).try(:output).try(:access_sas)
+
+          unless sas_url
+            msg = "Unable to find an SAS URL for #{disk_name}/#{resource_group}"
+            raise Azure::Armrest::NotFoundException.new(response.code, msg, response.body)
+          end
+
+          # The same restrictions that apply to the StorageAccont method also apply here.
+          range = byte_options[:range] if byte_options[:range]
+          range = byte_options[:start_byte]..byte_options[:end_byte] if byte_options[:end_byte]
```
bzwei:
```
range ||= byte_options[:start_byte]..byte_options[:end_byte] if byte_options[:start_byte] && byte_options[:end_byte]
```
same for the next line

lib/azure/armrest/storage/disk_service.rb
```diff
@@ -11,6 +11,112 @@ class DiskService < ResourceGroupBasedService
...
+        #   data = sds.get_blob_raw(disk.name, disk.resource_group, :range => 0..1023)
+        #
+        #   p data.headers
+        #   File.open('vm.vhd', 'a'){ |fh| fh.write(data.body) }
+        #
+        def get_blob_raw(disk_name, resource_group = configuration.resource_group, byte_options = {}, local_options = {})
+          validate_resource_group(resource_group)
+
+          local_options[:access] = 'read' # Must be 'read' at the moment
+          local_options[:durationInSeconds] ||= local_options[:duration_in_seconds] || 3600 # 1 hour default
```
bzwei: What are other available keys for `local_options`? Our users don't need to know how Azure API works; therefore we don't require them to provide the exact options (e.g., `durationInSeconds`) that accepted by the API. If there are no other required options I would have the method signature as
```
def get_blob_raw(disk_name, resource_group=..., options={})
```
The allowed keys in `options` will be the location parameters that you listed above, plus `:duration_in_seconds`

Issue #270, opened 2017-04-10
ArmrestCollection#to_json busted


djberg96:
If you try to convert an ArmrestCollection object to JSON using the `to_json` method, you get an error:

`vms.list_all.to_json`

Result:

```
lib/azure/armrest/model/base_model.rb:84:in `to_json': wrong number of arguments (given 1, expected 0) (ArgumentError)
```

This should "just work".

PR #271, opened 2017-04-10
Add an options argument to BaseModel#to_json

djberg96:
In order to maintain interface compatibility with existing to_json methods, we must add an options argument to the `to_json` method. Otherwise, attempting to call `to_json' on an array of model objects will fail.

Addresses https://github.com/ManageIQ/azure-armrest/issues/270

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -81,7 +81,10 @@ module Azure
         @hash
       end
 
-      def to_json
+      # Return the original JSON for the model object. The +options+ argument
+      # is for interface compatibility only.
+      #
+      def to_json(_options = nil)
         @json
       end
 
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/model/base_model.rb
```diff
@@ -81,7 +81,10 @@ def to_hash
         @hash
       end
 
-      def to_json
+      # Return the original JSON for the model object. The +options+ argument
+      # is for interface compatibility only.
+      #
+      def to_json(_options = {})
```
bzwei: default argument should be `nil`

Here is the doc
```
# File activesupport/lib/active_support/json/encoders/enumerable.rb, line 10
  def to_json(options = nil) #:nodoc:
    "[#{map { |value| ActiveSupport::JSON.encode(value, options) } * ','}]"
  end
```

djberg96: @bzwei ok, updated.

PR #273, opened 2017-04-18
Added some convenience methods to the VirtualMachine model

djberg96:
This just adds some convenient shortcut methods for the VirtualMachine model. They are:

* managed_disk?
* size (alias: flavor)
* operating_system (alias: os)

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -263,3 +263,4 @@ module Azure
 end
 
 require_relative 'storage_account'
+require_relative 'virtual_machine'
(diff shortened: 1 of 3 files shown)


lib/azure/armrest/model/base_model.rb
```diff
@@ -263,3 +263,4 @@ class Snapshot < BaseModel; end
 end
 
 require_relative 'storage_account'
+require_relative 'virtual_machine'
```
bzwei: why this line is needed?

djberg96: Because otherwise it's not loaded.

bzwei: How could the gem work before without this line? From the your code changes I cannot see the additional need. Was it the spec test?

djberg96: @bzwei It's already defined in base_model.rb, but it's just a stub really. We already did something similar with StorageAccount.

Issue #275, opened 2017-05-08
StorageAccountService#list_all_private_images results in Errno::ECONNREFUSED


djberg96:
On rare occasions I'm hitting an ECONNREFUSED error when attempting to use the `list_all_private_images` method. This ultimately comes from the `StorageAccount#blobs` method. I am not sure if this is an issue with the storage endpoint, or an issue with the Parallel gem and http requests.

https://github.com/ManageIQ/azure-armrest/blob/master/lib/azure/armrest/model/storage_account.rb#L277

```
/Users/djberge/.rbenv/versions/2.4.1/lib/ruby/2.4.0/net/http.rb:906:in `rescue in block in connect': Failed to open TCP connection to foo.blob.core.windows.net:443 (Connection refused - connect(2) for "foo.blob.core.windows.net" port 443) (Errno::ECONNREFUSED)
	from /Users/djberge/.rbenv/versions/2.4.1/lib/ruby/2.4.0/net/http.rb:903:in `block in connect'
	from /Users/djberge/.rbenv/versions/2.4.1/lib/ruby/2.4.0/timeout.rb:93:in `block in timeout'
	from /Users/djberge/.rbenv/versions/2.4.1/lib/ruby/2.4.0/timeout.rb:103:in `timeout'
	from /Users/djberge/.rbenv/versions/2.4.1/lib/ruby/2.4.0/net/http.rb:902:in `connect'
	from /Users/djberge/.rbenv/versions/2.4.1/lib/ruby/2.4.0/net/http.rb:887:in `do_start'
	from /Users/djberge/.rbenv/versions/2.4.1/lib/ruby/2.4.0/net/http.rb:876:in `start'
	from /Users/djberge/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/rest-client-2.0.2/lib/restclient/request.rb:715:in `transmit'
	from /Users/djberge/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/rest-client-2.0.2/lib/restclient/request.rb:145:in `execute'
	from /Users/djberge/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/rest-client-2.0.2/lib/restclient/request.rb:52:in `execute'
	from /Users/djberge/Dev/azure-armrest-djberg96/lib/azure/armrest/armrest_service.rb:200:in `rest_execute'
[...]

PR #276, opened 2017-05-09
Deal with failure when calling blob_properties inside get_private_images

djberg96:
This is an attempt to deal with Timeout errors that occasionally occur when calling `StorageAccount#list_all_private_images`. Specifically, it deals with timeouts (or ECONNREFUSED) in the call to `StorageAccount#blob_properties`.

https://github.com/ManageIQ/azure-armrest/issues/275 (second issue)

It catches the error, sleeps for 1 second, then retries. Up to a maximum of 3 attempts. If it still fails, it issues a warning and moves on instead of failing completely.

### DIFF
--- a/lib/azure/armrest/storage_account_service.rb
+++ b/lib/azure/armrest/storage_account_service.rb
@@ -292,7 +292,15 @@ module Azure
             next unless File.extname(blob.name).casecmp('.vhd') == 0
             next unless blob.properties.lease_state.casecmp('available') == 0
 
-            blob_properties = storage_account.blob_properties(blob.container, blob.name, key)
+            # In rare cases the endpoint will be unreachable. Warn and move on.
+            begin
+              blob_properties = storage_account.blob_properties(blob.container, blob.name, key)
+            rescue Errno::ECONNREFUSED, Azure::Armrest::TimeoutException => err
+              msg = "Unable to collect blob properties for #{blob.name}/#{blob.container}: #{err}"
+              log('warn', msg)
+              next
+            end
+
             next unless blob_properties.respond_to?(:x_ms_meta_microsoftazurecompute_osstate)
             next unless blob_properties.x_ms_meta_microsoftazurecompute_osstate.casecmp('generalized') == 0
 
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -274,7 +274,13 @@ module Azure
         mutex = Mutex.new
 
         Parallel.each(containers(key), :in_threads => max_threads) do |container|
-          mutex.synchronize { array.concat(blobs(container.name, key)) }
+          begin
+            mutex.synchronize { array.concat(blobs(container.name, key)) }
+          rescue Errno::ECONNREFUSED, Azure::Armrest::TimeoutException => err
+            msg "Unable to gather blob information for #{container.name}: #{err}"
+            log('warn', msg)
+            next
+          end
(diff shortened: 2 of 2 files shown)


lib/azure/armrest/storage_account_service.rb
```diff
@@ -292,7 +292,21 @@ def get_private_images(storage_accounts)
             next unless File.extname(blob.name).casecmp('.vhd') == 0
             next unless blob.properties.lease_state.casecmp('available') == 0
 
-            blob_properties = storage_account.blob_properties(blob.container, blob.name, key)
+            max_retries = 0
+
+            begin
+              blob_properties = storage_account.blob_properties(blob.container, blob.name, key)
+            rescue Errno::ECONNREFUSED, Azure::Armrest::TimeoutError => err
+              if max_retries < 3
```
bronaghs: its not really a max_retry though, right? Its more of a retry_count or retries

lib/azure/armrest/storage_account_service.rb
```diff
@@ -292,7 +292,21 @@ def get_private_images(storage_accounts)
...
+                max_retries += 1
+                sleep 1
```
bronaghs: @djberg96 - do you think 1 second is long enough to wait for a retry?

lib/azure/armrest/storage_account_service.rb
```diff
@@ -292,7 +292,21 @@ def get_private_images(storage_accounts)
...
+                retry
+              else
+                warn "WARNING: Unable to collect blob properties for #{blob.name}/#{blob.container}: #{err}"
```
bronaghs: @djberg96 - is this how we log in the gem?

bzwei:
Why particularly `blob_properties` method can timeout? Is it because the amount of data returned can be huge?
In general as a low level gem (that wraps around the API) it should leave error handling or retry to the app.

djberg96:
@bzwei I'm not actually sure why `blob_properties` is timing out sporadically. It could be some kind of threading issue on our end, or it could be that the endpoint is flaking out temporarily on the Azure side. I'm leaning towards the latter.

Generally, it's true that we should let the app handle the error. But, because of the nature of this method, with a method call being made inside a loop, the only way the app could handle it is "all or nothing". This way it just skips over a failure.

lib/azure/armrest/storage_account_service.rb
djberg96: Ok. :)

lib/azure/armrest/storage_account_service.rb
djberg96: It only seems to flake out very briefly in my experience, so yes. And, if not, it will try again.

lib/azure/armrest/storage_account_service.rb
djberg96: Yeah, I guess I should use the actual log there. We may have to revisit logging in the near future since there has been an enhancement to the rest-client gem, but that's another story.

bzwei:
I can agree to skip failed API calls in the context of looping (and log the failure), but not favor of a blind retry mechanism.

djberg96:
@bzwei @bronaghs Ok, I've removed the retry logic and the message is now sent to the log.

djberg96:
@jrafanie Looping you in since we've been discussing it on the core list. :)

djberg96:
@bronaghs @bzwei Also added guards in the `all_blobs` method.

PR #278, opened 2017-05-24
Add the StorageAccount#update_blob_properties method

djberg96:
This adds the `update_blob_properties` to the `StorageAccount` model. The main reason for adding this is to add the ability to resize a blob. This may be desired for older .vhd files that are no longer a "valid" size according to current Azure standards.

I've also updated the internal api version. Look here for a list of changes:

https://docs.microsoft.com/en-us/rest/api/storageservices/versioning-for-the-azure-storage-services

I lowercased and stringified the header information to be consistent with the azure-signature gem as well, which prevents potential duplicates.

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -35,7 +35,7 @@ module Azure
 
       def initialize(json)
         super
-        @storage_api_version = '2015-12-11'
+        @storage_api_version = '2016-05-31'
         @proxy = ENV['http_proxy']
         @ssl_version = 'TLSv1'
         @ssl_verify = nil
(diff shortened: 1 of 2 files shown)


djberg96:
@bzwei Look ok?

lib/azure/armrest/model/storage_account.rb
```diff
@@ -217,6 +217,48 @@ def blob_properties(container, blob, key = nil, options = {})
...
+      # to resize the blob.
+      #
+      def update_blob_properties(container, blob, key = nil, options = {})
+        key ||= properties.key1
+
+        url = File.join(properties.primary_endpoints.blob, container, blob) + "?comp=properties"
+
+        hash = {'verb' => 'PUT'}
+
+        options.each do |okey, ovalue|
```
bzwei: you can use `options.transform_keys!`

bzwei: actually prefer `options.transform_keys` and leave the original options untouched.

djberg96: Oh, I didn't know about that method. I'll definitely take a look.

djberg96: @bzwei Done!

lib/azure/armrest/model/storage_account.rb
```diff
@@ -217,6 +217,47 @@ def blob_properties(container, blob, key = nil, options = {})
...
+        response = ArmrestService.send(
+          :rest_put,
+          :url         => url,
+          :headers     => headers,
+          :proxy       => proxy,
+          :ssl_version => ssl_version,
+          :ssl_verify  => ssl_verify
+        )
+
+        BlobProperty.new(response.headers.merge(:container => container, :name => blob))
```
bzwei: Is it really needed or just a convenience to insert `:container` and `:name`?

djberg96: They're somewhat for convenience, but are necessary for use with other `StorageAccount` methods.

PR #279, opened 2017-05-24
Wrap the File Services API

djberg96:
At the moment we don't have any support for the File Services API. This adds the following methods to the `StorageAccount` model:

* create_directory
* delete_directory
* directory_properties
* directory_metadata
* create_file
* delete_file
* copy_file
* file_properties
* file_metadata
* files
* file_content
* add_file_content

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -20,6 +20,10 @@ module Azure
       class Table < BaseModel; end
       class TableData < BaseModel; end
 
+      # Classes used to wrap file shares
+      class ShareDirectory < BaseModel; end
+      class ShareFile < BaseModel; end
+
       # The version string used in headers sent as part any internal http
       # request. The default is 2016-05-31.
       attr_accessor :storage_api_version
@@ -50,7 +54,7 @@ module Azure
       def tables(key = access_key)
         raise ArgumentError, "No access key specified" unless key
         response = table_response(key, nil, "Tables")
-        JSON.parse(response.body)['value'].map{ |t| Table.new(t) }
+        JSON.parse(response.body)['value'].map { |t| Table.new(t) }
       end
 
       # Return information about a single table for the given storage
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/model/storage_account.rb
```diff
@@ -119,6 +123,295 @@ def table_data(name, key = access_key, options = {})
...
+      ### Files and Directories
+
+      # Create a new directory under the specified +share+ or parent directory.
+      #
+      # The only supported option at this time is a "timeout" option.
+      #
+      def create_directory(share, directory, key = access_key, options = {})
+        raise ArgumentError, "No access key specified" unless key
+
+        query = "restype=directory"
```
bzwei: Can you have key and value pairs in a hash, and call `hash.to_query`? It improves readability.

djberg96: Oh, didn't know about that method, I'll take a look.

djberg96:
@bzwei Ok, made your suggested changes.

lib/azure/armrest/model/storage_account.rb
```diff
@@ -119,6 +123,287 @@ def table_data(name, key = access_key, options = {})
...
+      # The only supported option at this time is a "timeout" option.
+      #
+      def create_directory(share, directory, key = access_key, options = {})
+        raise ArgumentError, "No access key specified" unless key
+
+        query = {:restype => 'directory'}.merge(options).to_query
+
+        response = file_response(key, query, 'put', '', File.join(share, directory))
+
+        headers = Azure::Armrest::ResponseHeaders.new(response.headers)
```
bzwei: These three lines have appear so many times. We can make them into a shared method
```
def response_headers(response)
  Azure::Armrest::ResponseHeaders.new(response.headers).tap do |headers|
    headers.response_code = response.code
  end
end
private response_headers
```

lib/azure/armrest/model/storage_account.rb
```diff
@@ -119,6 +123,287 @@ def table_data(name, key = access_key, options = {})
...
+        headers.response_code = response.code
+
+        headers
+      end
+
+      # Delete the specified +share+ or parent directory.
+      #
+      # The only supported option at this time is a "timeout" option.
+      #
+      def delete_directory(share, directory, key = access_key, options = {})
```
bzwei: extract the implementation of both methods into a common one, and pass in the operation `put` or `delete`

lib/azure/armrest/model/storage_account.rb
```diff
@@ -119,6 +123,287 @@ def table_data(name, key = access_key, options = {})
...
+        response = file_response(key, query, 'get', '', File.join(share, directory))
+
+        ShareDirectory.new(response.headers)
+      end
+
+      # Get metadata for the specified +share+ or parent directory.
+      #
+      # The only supported option at this time is a "timeout" option.
+      #
+      def directory_metadata(share, directory, key = access_key, options = {})
```
bzwei: `directory_properties` and `directory_metadata` can share implementation

lib/azure/armrest/model/storage_account.rb
```diff
@@ -119,6 +123,287 @@ def table_data(name, key = access_key, options = {})
...
+
+        response = file_response(key, query, 'get', '', File.join(share, file))
+        response.body
+      end
+
+      # Returns the raw contents of the specified file.
+      #
+      # The only supported option at this time is a "timeout" option.
+      #
+      def file_properties(share, file, key = access_key, options = {})
```
bzwei: `file_content` and `file_properties` share same implementation

lib/azure/armrest/model/storage_account.rb
```diff
@@ -119,6 +123,287 @@ def table_data(name, key = access_key, options = {})
...
+
+        headers = Azure::Armrest::ResponseHeaders.new(response.headers)
+        headers.response_code = response.code
+
+        headers
+      end
+
+      # Copy a +src_file+ to a destination +dst_file+ within the same storage account.
+      #
+      def copy_file(src_container, src_file, dst_container = src_container, dst_file = nil, key = access_key)
```
bzwei: `copy_file` and `create_file` share implementation

lib/azure/armrest/model/storage_account.rb
djberg96: I'm not going to create an abstract method that only works for 2 methods. It has to be 3 or more for me to consider it worthwhile. Otherwise it's just unncessary abstraction that makes the code harder to follow.

lib/azure/armrest/model/storage_account.rb
djberg96: I'm happy to switch to using `tap`, though I don't think it warrants its own method.

djberg96:
@bzwei I've update the returns to use `tap` but that's all I want to do for now. I would like to merge this and then release 0.8.0 because I want to integrate some of the existing changes into ManageIQ. We can always refactor later.

bzwei:
@djberg96 you may want to open an issue for the refactoring

PR #280, opened 2017-06-08
Explicitly add the resource_group for private images

djberg96:
At the moment the custom objects we generate for the `list_private_images` method do not include a resource group. This PR explicitly sets it, using the underlying storage account's resource group as its source.

### DIFF
--- a/lib/azure/armrest/storage_account_service.rb
+++ b/lib/azure/armrest/storage_account_service.rb
@@ -328,7 +328,7 @@ module Azure
           )
         )
 
-        StorageAccount::PrivateImage.new(hash)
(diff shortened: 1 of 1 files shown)


lib/azure/armrest/storage_account_service.rb
```diff
@@ -328,7 +328,10 @@ def blob_to_private_image_object(storage_account, blob, blob_properties)
           )
         )
 
-        StorageAccount::PrivateImage.new(hash)
+        image = StorageAccount::PrivateImage.new(hash)
```
bzwei: `StorageAccount::PrivateImage.new(hash).tap { |image| image.resource_group = storage_account.resource_group }`

djberg96: Ok, done.

PR #282, opened 2017-06-13
Add predefined Environments for Germany and China

djberg96:
This PR adds Environment objects for both China and Germany. This will let us support more Azure regions in the ManageIQ app as well.

I also added some specs.

### DIFF
--- a/lib/azure/armrest/environment.rb
+++ b/lib/azure/armrest/environment.rb
@@ -86,7 +86,7 @@ module Azure
 
       # Pre-generated environments
 
-      Public = self.new(
+      Public = new(
         :name                          => 'Public',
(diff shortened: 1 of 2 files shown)


bzwei:
shouldn't them be all capital, e.g., GERMANY and CHINA? Also the rubocop complaints about RedundantSelf.

djberg96:
@bzwei No, the others aren't in all-caps, so I didn't make these all caps, since they're actually instances.

I also removed the `self` that rubocop was complaining about.

bzwei:
I think all caps are for constants while capitalized are for class or module names. For existing constants can we change to all caps? Maybe I am wrong.

djberg96:
@bzwei I think that ship has sailed. And there were no complaints from rubocop.

PR #284, opened 2017-06-15
Remove check for key1, add access_key accessor

djberg96:
When we first wrote the StorageAccount model, the Azure API included `key1` and `key2` in the output when getting information for individual storage accounts. Consequently, the `key` argument for most of the methods defaulted to `properties.key1` if a key wasn't explicitly provided.

However, Microsoft later removed this information from the JSON output, requiring a separate request to get access key information. The oldest supported api-version string (2015-05-01-preview) does not include this output, so there is no point keeping it around. On the contrary, if you don't provide a key explicitly, you will get a cryptic error message about there being no "key1" property.

This PR does three things. First, it adds an `access_key` accessor, so that you can set it once for that particular storage account. Second, it changes the default from `nil` to `access_key` for all of the method signatures. Lastly, it performs a check against the key argument. If both the argument and the accessor are still `nil` then an ArgumentError is raised.

This change is, for all intents and purposes, backwards compatible because currently you must still explicitly provide a key or that cryptic error is raised.

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -33,6 +33,9 @@ module Azure
       # The SSL verification method used for each request. The default is VERIFY_PEER.
       attr_accessor :ssl_verify
 
+      # The default access key used when creating a signature for internal http requests.
+      attr_accessor :access_key
+
       def initialize(json)
         super
         @storage_api_version = '2016-05-31'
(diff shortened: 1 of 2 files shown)


djberg96:
@bzwei Look good?

bzwei:
@djberg96 do we really need to `check_for_key`?

djberg96:
@bzwei It's not strictly necessary, no.

bzwei:
Should we remove it then? The code will be cleaner.

PR #287, opened 2017-07-07
Refactor the locations method

djberg96:
The ArmrestService#locations method originally iterated over the providers list because I didn't think there was a direct way to get a list of supported locations. The only documentation I could find applied only to classic Azure. However, it turns out that you *can* get a list of locations directly.

Consequently I would like to refactor the ArmrestService#locations method so that it returns a list of Location objects. Also, to be consistent with our other methods, change the method name to `list_locations` with a deprecated alias for backwards compatibility. Here's an example of what the proposed method returns:

    "id"          => "/subscriptions/xxx/locations/koreasouth",
    "name"        => "koreasouth",
    "displayName" => "Korea South",
    "longitude"   => "129.0756",
    "latitude"    => "35.1796"

Getting locations by provider can still be accomplished by using the `ResourceProviderService` if users desire.

Since this is a backwards incompatible change, it should be part of the 0.8.0 release, along with my proposed changes to `StorageAccount#create_blob` and `StorageAccount#create_blob_snapshot`.

FYI, this does cause some minor breakage in ManageIQ for discovery, but the fix is very simple (a modification to 1 line), and we wouldn't have to worry about it until we upgraded to 0.8.x of the armrest gem.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -77,15 +77,25 @@ module Azure
       # +provider+. If you do not specify a provider, then the locations for
       # all providers will be returned.
       #
-      # If you need individual details on a per-provider basis, use the
-      # provider_info method instead.
-      #--
+      # If you need individual details on a per-provider basis, use the methods
+      # of the ResourceProviderService instead.
+      #
+      # Deprecated.
       #
       def locations(provider = nil)
         list = configuration.providers
         list = list.select { |rp| rp.namespace.casecmp(provider) == 0 } if provider
-
         list.collect { |rp| rp.resource_types.map(&:locations) }.flatten.uniq.sort
+      end
+
+      deprecate :locations, :list_locations, 2019, 1
+
+      # Returns a list of Location objects for the current subscription.
+      #
+      def list_locations
+        url = url_with_api_version(configuration.api_version, base_url, 'locations')
+        response = rest_get(url)
+        Azure::Armrest::ArmrestCollection.create_from_response(response, Location)
       end
 
       # Returns a list of subscriptions for the current tenant.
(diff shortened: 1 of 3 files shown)


bzwei:
Since this is not a backward compatible change, how about keeping `locations` as is but mark it deprecated? We then introduce the new `list_locations` method.

djberg96:
@bzwei I think having `locations` and `list_locations` have different behavior would be very confusing.

bzwei:
@djberg96 I don't think it is confusing if the old `locations` method returns an array of strings (locations) which can be considered as a simple method; while the new `list_locations` lists the detailed locations.

djberg96:
@bzwei Ok, updated.

lib/azure/armrest/armrest_service.rb
```diff
@@ -77,17 +77,27 @@ def get_provider(provider)
...
+      # Deprecated.
       #
       def locations(provider = nil)
         list = configuration.providers
         list = list.select { |rp| rp.namespace.casecmp(provider) == 0 } if provider
-
         list.collect { |rp| rp.resource_types.map(&:locations) }.flatten.uniq.sort
       end
 
+      deprecate :locations, :list_locations, 2019, 1
```
bzwei: Will we deprecate `list_locations` too?

spec/armrest_service_spec.rb
```diff
@@ -17,10 +17,15 @@
   end
 
   context "instance methods" do
-    it "defines a locations method" do
+    it "defines a list method" do
```
bzwei: revert this change.

lib/azure/armrest/armrest_service.rb
djberg96: Hm? No.

spec/armrest_service_spec.rb
djberg96: Updated.

lib/azure/armrest/armrest_service.rb
bzwei: Sorry I mis-read the deprecate statement

PR #290, opened 2017-07-17
Add list_all method to VirtualMachineImageService

01100010011001010110010101110000:
Essentially just the commit message. This adds a list_all method to the ```VirtualMachineImageService``` class. This method is used in ManageIQ/manageiq-providers-azure#95


djberg96:
Looks good to me. Any chance you could add some basic specs? Doesn't have to be anything fancy.

01100010011001010110010101110000:
To be honest, I don't know a great deal about writing specs. If there is more that I missed, let me know

PR #292, opened 2017-08-02
Fix start_byte + length options in DiskService

djberg96:
The `DiskService#get_blob_raw` method has a bug when :start_byte and :length are used together. This fixes that.


djberg96:
@jerryk55 please double check my logic

jerryk55:
Yup that's basically the way I coded it in my local copy of the gem.  LGTM @djberg96 !  👍

PR #293, opened 2017-08-07
Add TemplateDeploymentService#exists? method

djberg96:
This adds the `exists?` method to the `TemplateDeploymentService` class. Ideally we could use this method to speed up template collection, as well as reduce memory, since it uses a HEAD request.


djberg96:
I'm deliberately using `and` here, so ignore rubocop.

PR #295, opened 2017-08-09
Add filter to TemplateDeploymentService#list_all

djberg96:
This adds the option of a filter to the `TemplateDeployment#list_all` option, which in turn modifies the underlying `list_in_all_groups` private method. Applying a filter up front could reduce the time and memory currently consumed by collecting deployment information.

This is part of the effort to help fix https://bugzilla.redhat.com/show_bug.cgi?id=1431912.

### DIFF
--- a/lib/azure/armrest/resource_group_based_service.rb
+++ b/lib/azure/armrest/resource_group_based_service.rb
@@ -1,3 +1,5 @@
+require 'active_support/core_ext/hash/conversions'
+
 module Azure
   module Armrest
     # Base class for services that need to run in a resource group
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -233,15 +235,31 @@ def validate_resource(name)
...
-        id_string = File.join(id_string, *args) unless args.empty?
-        id_string
+
+        query = "?api-version=#{@api_version}"
+
+        args.each do |arg|
+          if arg.kind_of?(Hash)
+            arg.each do |key, value|
+              if key.to_s.casecmp('top').zero?
+                query << "&$top=#{value}"
```
bronaghs: @djberg96 - what does 'top' mean?

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -233,15 +235,31 @@ def validate_resource(name)
...
+              else
+                query << "&$filter=" unless query.include?("$filter")
```
bronaghs: @djberg96 - Wont query only contain the api-version at this point?

lib/azure/armrest/resource_group_based_service.rb
djberg96: It's a limit like "first", e.g. "$top=3" would return the first 3 results.

lib/azure/armrest/resource_group_based_service.rb
djberg96: Not if they submit multiple filters. Though, come to think of it, that wouldn't actually work at the moment because we need an "and".

lib/azure/armrest/resource_group_based_service.rb
bronaghs: ah ok

djberg96:
@bronaghs Ok, updated to handle multiple filters and a raw filter if a user wants to specify their own.

djberg96:
I'm ignoring those rubocop warnings btw.

PR #297, opened 2017-08-10
Add the ResourceGroupService#exists? method

djberg96:
Adds the ResourceGroupService#exists? method. Unlike a `get` this is a faster HEAD request, and only returns a boolean value.


djberg96:
Ignore rubocop, that code is intentional

PR #299, opened 2017-08-18
Add get_blob_raw for Managed Disk Snapshots

jerryk55:
Add ability to read Snapshots taken for Managed Disks.
We add get_blob_raw for the Snapshots which mirrors the code for
Managed Disks themselves via the Storage DiskService.
A common helper file raw_blob_helper.rb with the RawBlobHelper
module is added to be used by the DiskService and SnapshotService.

In addition the EndGetAccess operation must be performed when each read
operation is finished so that an attempt to remove the Snapshot will succeed.

These changes as discussed with @djberg96.  @djberg96, @roliveri please review and
merge when appropriate.  Thanks.  Other PRs for manageiq-providers-azure  that depends
on this one and manageiq  are forthcoming.

The following BZ depends on this change: https://bugzilla.redhat.com/show_bug.cgi?id=1459612

### DIFF
--- a/lib/azure/armrest/storage/disk_service.rb
+++ b/lib/azure/armrest/storage/disk_service.rb
@@ -6,8 +6,8 @@ module Azure
     module Storage
       # Base class for managing disks.
       class DiskService < ResourceGroupBasedService
-        require_relative './raw_blob_helper'
-        include Azure::Armrest::Storage::RawBlobHelper
+        require_relative 'managed_storage_helper'
+        include Azure::Armrest::Storage::ManagedStorageHelper
 
         # Create and return a new DiskService instance.
         #
--- a/lib/azure/armrest/storage/disk_service.rb
+++ b/lib/azure/armrest/storage/disk_service.rb
@@ -6,124 +6,13 @@ module Azure
     module Storage
       # Base class for managing disks.
       class DiskService < ResourceGroupBasedService
+        require_relative './raw_blob_helper'
+        include Azure::Armrest::Storage::RawBlobHelper
+
         # Create and return a new DiskService instance.
         #
         def initialize(configuration, options = {})
           super(configuration, 'disks', 'Microsoft.Compute', options)
-        end
-
-        # Get the raw blob information for a managed disk. This is similar to
-        # the StorageAccount#get_blob_raw method, but applies only to a managed
-        # disk, whereas that method applies only to an individual storage
-        # account.
-        #
-        # As with the Storage#get_blob_raw method, you should pass a :range,
-        # :start_byte, :end_byte or :length option. If you want the entire
-        # image you must pass the :entire_image option, though this is generally
-        # not recommended. Unlike the Storage#get_blob_raw method, this method
-        # does not support the :date parameter.
-        #
-        # The +options+ are as follows:
-        #
-        #   :range        => A range of bytes you want, e.g. 0..1023 to get first 1k bytes
-        #   :start_byte   => The starting byte number that you want to collect bytes for. Use
-        #                    this in conjunction with :length or :end_byte.
-        #   :end_byte     => The ending byte that you want to collect bytes for. Use this
-        #                    in conjunction with :start_byte.
-        #   :length       => If given a :start_byte, specifies the number of bytes from the
-        #                    the :start_byte that you wish to collect.
-        #   :entire_image => If set, returns the entire image in bytes. This will be a long
-        #                    running request that returns a large number of bytes.
-        #
-        # You may also pass a :duration parameter, which indicates how long, in
-        # seconds, that the privately generated SAS token should last. This token
-        # is used internally by requests that are used to access the requested
-        # information. By default it lasts for 1 hour.
-        #
-        # Get the information you need using:
-        #
-        # * response.body    - blob data (the raw bytes).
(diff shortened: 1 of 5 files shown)


lib/azure/armrest/storage/disk_service.rb
```diff
@@ -6,125 +6,14 @@ module Armrest
     module Storage
       # Base class for managing disks.
       class DiskService < ResourceGroupBasedService
+        require_relative './raw_blob_helper'
```
djberg96: You shouldn't need the './' here.

lib/azure/armrest/storage/raw_blob_helper.rb
```diff
@@ -0,0 +1,123 @@
+module Azure::Armrest::Storage::RawBlobHelper
```
djberg96: Maybe instead of "RawBlobHelper" we just call it "ManagedStorageHelper" in case we decide to add any other common set of methods.

lib/azure/armrest/storage/disk_service.rb
jerryk55: Right.  Fixing.

lib/azure/armrest/storage/raw_blob_helper.rb
jerryk55: Sounds better.  Changing.

jerryk55:
@djberg96 your comments have been addressed as well as (most) Rubocop warnings.

jerryk55:
Forgot to update the spec tests for the SnapshotService to add get_blob_raw.  Done now.  (lets see it pass please)

lib/azure/armrest/storage/managed_storage_helper.rb
```diff
@@ -0,0 +1,123 @@
...
+          :url         => sas_url,
+          :headers     => headers,
+          :proxy       => configuration.proxy,
+          :ssl_version => configuration.ssl_version,
+          :ssl_verify  => configuration.ssl_verify
+        )
+      rescue RestClient::Exception, Azure::Armrest::ForbiddenException => err
+        retries += 1
+        raise err unless retries < max_retries
+        log("get_blob_raw: 403 Forbidden received on read - retry number #{retries}")
```
djberg96: Instead of hard coding this message, you should be able to get the actual message from the `err` object.

lib/azure/armrest/storage/managed_storage_helper.rb
```diff
@@ -0,0 +1,123 @@
...
+        retry
+      end
+    ensure
+      end_url = build_url(resource_group, disk_name, 'EndGetAccess')
```
djberg96: Maybe this should be wrapped in an "`f retries <= 5` clause in case it never succeeded in the first place.

djberg96:
@jerryk55 mostly looks good, found a couple other things.

lib/azure/armrest/storage/managed_storage_helper.rb
jerryk55: Then I would do something like:
log('info', "get_blob_raw: #{err.to_s} - retry number #{retries}") 
sound good?

lib/azure/armrest/storage/managed_storage_helper.rb
jerryk55: @djberg96 I'm not sure what you'd solve with that.  The retry is only for the failure of the actual read of the disk.  If anything prior to that (for instance the 'BeginGetAccess' call) fails for any reason we will hit the ensure clause but no retries will happen.

lib/azure/armrest/storage/managed_storage_helper.rb
djberg96: No need to call `.to_s` since it's being interpolated already, but yeah, that should work. And 'info' should probably be 'error' there.

lib/azure/armrest/storage/managed_storage_helper.rb
djberg96: I just mean that I'm not sure what will happen on a call to `EndGetAccess` if `BeginGetAccess` never worked. It will probably be a no-op I suppose, so perhaps I'm worrying about nothing.

lib/azure/armrest/storage/managed_storage_helper.rb
jerryk55: When we retry successfully the message shouldn't be an error - we are informing that the retry is occurring.  The exception when we exceed the number of allowable retries will be the error case.

lib/azure/armrest/storage/managed_storage_helper.rb
jerryk55: That was my assumption too.

lib/azure/armrest/storage/managed_storage_helper.rb
djberg96: True, I guess `info` or `warn` would be fine.

djberg96:
@jerryk55 looks good except for that hard coded message. Once that's updated I'll merge this.

jerryk55:
Message updated to "WARN" level and the "err" value from the exception is now used.  @djberg96 this should be good to go.  Thanks!

jerryk55:
@miq-bot add_label fine/yes

PR #301, opened 2017-09-07
Modify the StorageAccountService class to pass full configuration object to model

djberg96:
Currently we are only passing specific parameters to each StorageAccount object from the Configuration object. This PR modifies the StorageAccountService and StorageAccount model classes so that the full configuration object is passed along.

This eliminates the need for separate accessors for each property, and makes it easier to handle configuration properties that we may need to handle in the future within the model. For example, I want to add a `:timeout` property to the Configuration object that I will need to pass along to the models. With this approach, the model will automatically have access to it.

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -28,24 +28,15 @@ module Azure
       # request. The default is 2016-05-31.
       attr_accessor :storage_api_version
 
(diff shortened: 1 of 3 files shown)


djberg96:
Note that using `ObjectSpace.memsize_of(account)`, the underlying size of the storage object did not change. It was 120 either way.

PR #302, opened 2017-09-08
Add a :timeout option that is passed along to all http requests

djberg96:
This PR adds a :timeout option that is passed along to each rest-client http request. I've set the default value to 10 seconds for now.

I've tested several of our methods, including private image collection and raw blob data, and it seems to be fine with that value. It can always be configured, too.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -288,6 +288,7 @@ module Azure
           :proxy       => configuration.proxy,
           :ssl_version => configuration.ssl_version,
           :ssl_verify  => configuration.ssl_verify,
+          :timeout     => configuration.timeout,
           :headers => {
             :accept        => configuration.accept,
             :content_type  => configuration.content_type,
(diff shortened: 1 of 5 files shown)


bzwei:
What should be the default? not setting timeout like before or 10 seconds? What's the problem without timeout being set?

djberg96:
@bzwei rest-client doesn't set a default, so it should default to whatever Net::HTTP uses, which I think is 60 seconds. That's too high IMO.

djberg96:
Marking as WIP for now, want to think about the default some more, and need to rebase anyway.

djberg96:
@bzwei In addition to the default being to high (though perhaps 10 is too low), adding a timeout option would also be something that could be configured on the Azure provider side via config/settings.yml.

bzwei:
If the restclient default is 60, we should set 60 here also, and allow caller to override. In general we should set the default value large enough.

djberg96:
@bzwei ok, I've changed the default to 60.

PR #303, opened 2017-09-08
Fix message bug

djberg96:
Seems I forgot an assignment operator.


bronaghs:
So @djberg96  this could explain why we didnt see this message in the logs we discussed yesterday.

djberg96:
@bronaghs it might!

djberg96:
@bronaghs i should add that i would expect to see an exception regarding an "undefined method 'msg'" or something along those lines then.

bronaghs:
True. I will wade through the logs again looking for it. If found, it would verify the timeout was rescued and could move the investigation forward.

PR #308, opened 2017-09-15
Add max_retries accessor and retry on rest_execute in certain conditions

djberg96:
This adds the Configuration#max_retries accessor. This, in turn, controls the maximum number of retry attempts on 429 and most 50x errors.

The `rest_execute` method is then modified to automatically retry X number of times (default: 3) if a 429 or 50x HTTP error code occurs. The 429 error is returned if the user exceeds their rate limit. The 50x errors that I rescue are server side issues that are typically temporary and transient.

For both 429 and 503 errors the server will return a `Retry-After` header that we can use to determine the amount of time we should sleep between tries. If that header isn't found, it defaults to 30 seconds. If a logger object is set, a warning is logged as well.

Update: I've added 409 (Conflict) as well since it can happen with blob operations.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -205,12 +205,24 @@ module Azure
       class << self
         private
 
-        def rest_execute(options, http_method = :get, encode = true)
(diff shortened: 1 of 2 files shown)


bzwei:
@djberg96 Good Job!

PR #309, opened 2017-09-19
Refactor get_blob_raw method

djberg96:
This revamps the get_blob_raw method to use a proper `wait` approach on the `BeginGetAccess` call, and bails after two minutes if unsuccessful. The various `try` blocks are unnecessary now, since they were an artifact of trying to reference properties before we actually succeeded in getting access.

This also updates it so that the `EndGetAccess` call is only made if the `BeginGetAccess` call was successful.

In addition, it modifies the `wait` method to accept a `wait_interval` argument (default 10 seconds), and updates the specs.

### DIFF
--- a/lib/azure/armrest/storage/managed_storage_helper.rb
+++ b/lib/azure/armrest/storage/managed_storage_helper.rb
@@ -58,64 +58,64 @@ module Azure::Armrest::Storage::ManagedStorageHelper
     }
 
     # This call will give us an operations URL in the headers.
-    initial_url = build_url(resource_group, disk_name, 'BeginGetAccess')
-    response    = rest_post(initial_url, post_options.to_json)
+    begin_get_access_url = build_url(resource_group, disk_name, 'BeginGetAccess')
+    begin_get_access_response = rest_post(begin_get_access_url, post_options.to_json)
+
+    headers = Azure::Armrest::ResponseHeaders.new(begin_get_access_response.headers)
+    status = wait(headers, 120, 5)
+
+    unless status.casecmp('succeeded').zero?
+      msg = "Unable to obtain an operations URL for #{disk_name}/#{resource_group}"
+      log('debug', "#{msg}: #{begin_get_access_response.headers}")
+      raise Azure::Armrest::NotFoundException.new(begin_get_access_response.code, msg, begin_get_access_response.body)
+    end
+
+    # Get the SAS URL from the BeginGetAccess call
+    op_url = headers.try(:azure_asyncoperation) || headers.location
+
+    # Dig the URL + SAS token URL out of the response
+    response = rest_get(op_url)
+
+    body = Azure::Armrest::ResponseBody.new(response.body)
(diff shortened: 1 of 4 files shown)


chessbyte:
For readability, I would take https://github.com/djberg96/azure-armrest/blob/5d20c31af8df7cea3fa6c86ff0151bfdde696735/lib/azure/armrest/storage/managed_storage_helper.rb#L60-L101 and extract it all to a method named `get_sas_url` or some other meaningful name.

bronaghs:
@jerryk55 @bzwei - can you please take a look?

djberg96:
@chessbyte @bronaghs @bzwei We actually have builtin `wait` and `poll` methods, and may have been silly by not using them. I got a bit confused because of the multi step nature of this method.

I've got a cleaner version over at https://github.com/djberg96/azure-armrest/commit/1436d0998070723b49e5e0143f607985fd4e6d30 that I'm having @jerryk55 test.

jerryk55:
FWIW this has been successfully tested in the customer? lab in AU/NZ that was exhibiting the issues we are addressing with  high profile BZ https://bugzilla.redhat.com/show_bug.cgi?id=1488967

lib/azure/armrest/storage/managed_storage_helper.rb
```diff
@@ -116,8 +115,10 @@ def get_blob_raw(disk_name, resource_group = configuration.resource_group, optio
         retry
       end
     ensure
-      end_url = build_url(resource_group, disk_name, 'EndGetAccess')
-      rest_post(end_url)
+      if begin_get_access_response
```
bzwei: In what case `begin_get_access_response` is `nil`?

jerryk55: Then we can't call EndGetAccess because BeginGetAccess didn't succeed.

djberg96: rest_post will return nil if an exception is raised. Though, I just realized that our ensure block is not on the method, but on the first begin block. Fixing...

bzwei:
The method now becomes quite lengthy. Consider splitting into smaller methods.

In line https://github.com/djberg96/azure-armrest/blob/d6cec3d24bb570e1da2b65ce12641fe1d0fff839/lib/azure/armrest/storage/managed_storage_helper.rb#L114 don't you need to pause a while before retry?

The `wait` method simply pulls/waits for an async POST to complete, right?  I don't understand your statement "we will make one more attempt after sleeping Retry-After seconds (or 30), to get the token". Where is the one more attempt? Do you mean wait on previous call or make another call?

djberg96:
@bzwei I've updated the description, which didn't make sense with regards to the current approach.

jerryk55:
This LGTM.

djberg96:
Removed WIP tag, I think it's ready.

@bzwei Agreed that this could be refactored and split up, but later.

PR #312, opened 2017-09-20
Retry Getting SAS URL Token

jerryk55:
After calling BeginGetAccess the SAS URL Token may not be immediately available.
Retry several times for up to 100 seconds finally raising the previously returned exception
if it fails to be acquired.  Note that we use the response.headers[:retry_after] value or a default of 5 seconds for the sleep.  Larger default sleep values slow down turnaround time too much.

Also do not call EndGetAccess unless the BeginGetAccess was successful.

This is in support of BZ https://bugzilla.redhat.com/show_bug.cgi?id=1475540
and the downstream clone https://bugzilla.redhat.com/show_bug.cgi?id=1488967

These is a high-priority issue and we need to review, merge, back port to Fine,
and create a hot fix as soon as we can.

@djberg96 @bronaghs @roliveri please review.
Tested successfully at the customer location exhibiting the issue.

### DIFF
--- a/lib/azure/armrest/storage/managed_storage_helper.rb
+++ b/lib/azure/armrest/storage/managed_storage_helper.rb
@@ -58,10 +58,11 @@ module Azure::Armrest::Storage::ManagedStorageHelper
     }
 
     # This call will give us an operations URL in the headers.
-    initial_url = build_url(resource_group, disk_name, 'BeginGetAccess')
-    response    = rest_post(initial_url, post_options.to_json)
+    begin_get_access_url      = build_url(resource_group, disk_name, 'BeginGetAccess')
+    begin_get_access_response = rest_post(begin_get_access_url, post_options.to_json)
+
     begin
-      headers = Azure::Armrest::ResponseHeaders.new(response.headers)
+      headers = Azure::Armrest::ResponseHeaders.new(begin_get_access_response.headers)
 
       # Using the URL returned from the above call, make another call that
       # will return the URL + SAS token.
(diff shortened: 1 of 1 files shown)


lib/azure/armrest/storage/managed_storage_helper.rb
```diff
@@ -58,28 +58,46 @@ def get_blob_raw(disk_name, resource_group = configuration.resource_group, optio
...
       unless op_url
         msg = "Unable to find an operations URL for #{disk_name}/#{resource_group}"
-        raise Azure::Armrest::NotFoundException.new(response.code, msg, response.body)
+        log('debug', "#{msg}: #{begin_get_access_response.headers}")
+        raise Azure::Armrest::NotFoundException.new(begin_get_access_response.code, msg, begin_get_access_response.body)
       end
 
       # Dig the URL + SAS token URL out of the response
       response = rest_get(op_url)
       body     = Azure::Armrest::ResponseBody.new(response.body)
```
djberg96: Jerry, couldn't we check the `body.status` here for "InProgress", and just loop until it's "Succeeded"?

lib/azure/armrest/storage/managed_storage_helper.rb
```diff
@@ -58,28 +58,46 @@ def get_blob_raw(disk_name, resource_group = configuration.resource_group, optio
...
       sas_url  = body.try(:properties).try(:output).try(:access_sas)
 
-      unless sas_url
-        msg = "Unable to find an SAS URL for #{disk_name}/#{resource_group}"
-        raise Azure::Armrest::NotFoundException.new(response.code, msg, response.body)
+      max_sas_url_retries = 20
+      sas_url_retries     = 0
+
```
chessbyte: this whole loop looks overly complicated and hard to read.  Can we extract it to a new method called `wait_for_async_operation` where all the trying and retrying is done?

lib/azure/armrest/storage/managed_storage_helper.rb
jerryk55: @djberg96 perhaps but I'm not sure if there is a 1-1 correlation between that status and the appearance of the  SAS URL token - do you know that for sure?

lib/azure/armrest/storage/managed_storage_helper.rb
jerryk55: @chessbyte let me discuss with @djberg96 regarding his previous comment and see what we can do to either simplify or extract as you suggest.

jerryk55:
Closing this PR in lieu of very similar https://github.com/ManageIQ/azure-armrest/pull/309 that I am working on with @djberg96.

PR #315, opened 2017-09-21
Use filter from list_all (remove duplicate code)

NickLaMuro:
The `ResourceGroupBasedService#list_all` method, which the `StorageAccountService` inherits from, has the exact same filter code in place as what is used in the `list_all_private_images` method, so use that instead of maintaining duplicate versions of this code.

Found this while poking around in the gem, so thought I would make a quick PR to address.  All of the tests seem to pass with these changes in place.  Will do some testing where I know this is used to also confirm the same functionality.

### DIFF
--- a/lib/azure/armrest/storage_account_service.rb
+++ b/lib/azure/armrest/storage_account_service.rb
@@ -172,16 +172,7 @@ module Azure
       # Note that for string values the comparison is caseless.
       #
       def list_all_private_images(filter = {})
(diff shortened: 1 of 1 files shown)


djberg96:
Nice catch. Will merge when you say it's ok.

NickLaMuro:
@djberg96 ran this a few times on my machine and it doesn't seem to change any results, or cause errors, so I would say I am as confident with this as I can reasonably be.  :crossed_fingers:

PR #318, opened 2017-09-22
Use ArmrestCollection#next_link in get_all_results

NickLaMuro:
Currently, when the `ArmrestService#get_all_results method` is called, it parses the response body (as JSON) once when instantiating the ArmrestCollection, and a second time to determine the `'nextLink'` value for the subsequent `while` loop to get the remaining results (this is duplicated as well for each iteration of the loop).

Since we now have a `next_link` variable on the `ArmrestCollection` instances that is determined in the same fashion from the first `JSON.parse`, we can use that instead of duplicating the same effort twice.

Benchmarks
----------

In practice, this halved the allocations caused by the JSON gem (when using the one from ruby stdlib), and 10% of the total object allocations over all when calling `StorageAccountService#list_all_private_images` on a small to moderate collection of `StorageAccounts` (around 200 in this instance, filtered by location to around 70).  CPU improvements might be observed on larger collections.

A slight uptick on memory is observed in the armrest codebase because of this caching, but it is minimal compared to the savings.

**Before**

```
Total allocated: 13192688 bytes (149972 objects)
Total retained:  35707 bytes (344 objects)

allocated memory by gem
   6702920  azure-armrest/lib
   3494509  activesupport/lib
   2279476  json-2.0.4
    330191  addressable-2.4.0
    244959  rest-client-2.0.2
...
```

**After**

```
Total allocated: 12062224 bytes (133571 objects)
Total retained:  35925 bytes (347 objects)

allocated memory by gem
   6711976  azure-armrest/lib
   3494509  activesupport/lib
   1139738  json-2.0.4
    330191  addressable-2.4.0
    245177  rest-client-2.0.2
...
```

### DIFF
--- a/lib/azure/armrest/armrest_collection.rb
+++ b/lib/azure/armrest/armrest_collection.rb
@@ -3,6 +3,7 @@
 module Azure
   module Armrest
     class ArmrestCollection < Array
+      attr_accessor :next_link
       attr_accessor :continuation_token
       attr_accessor :response_headers
       attr_accessor :response_code
@@ -21,7 +22,8 @@ module Azure
 
           array.response_code = response.code
           array.response_headers = response.headers
-          array.continuation_token = parse_skip_token(json_response)
+          array.next_link = json_response['nextLink']
+          array.continuation_token = parse_skip_token(array.next_link)
 
           array
         end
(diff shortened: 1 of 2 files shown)


djberg96:
I like it, thanks.

PR #319, opened 2017-09-22
Light base model mode

NickLaMuro:
The intent of this PR is to introduce a flag into the initializer of each `BaseModel` backed class, currently called `skip_accessors_definition`.

The intent of this is that it can be used internally to avoid calling `__setobj__` on objects initially, which is useful for methods like `StorageAccountService#list_all_private_images`, which uses multiple intermediate objects/api calls to get the end result, and calling __setobj__ on each one of those is expensive and resource intensive, especially when they don't need to be used by the end user.

Some helper methods also needed to be introduced to allow for safe access of properties that were in nested hashes.  To save on resources, I avoided using `.with_indifferent_access` to the `@hash` value of the `BaseModel`, but that might have made some of those methods unecessary.  For now, they at least show where these methods were being used.

Benchmarks
----------
When running `StorageAccountService#list_all_private_images` on a small to moderate collection of StorageAccounts (around 200 in this instance, filtered by location to around 70), the memory dropped by over 50%, and as well as the execution time.

_**Note**: "Before" was tested using the new `0.8.3` release of the gem.  These benchmarks were a bit slower prior to that release.  See links section for included performance PRs that made it into the `0.8.3` release._

**Before**

```
Mem: 897.89453125MiB

real    0m50.757s
user    0m49.354s
sys     0m1.207s
```

**After**

```
Mem: 389.14453125MiB

real    0m31.342s
user    0m30.177s
sys     0m0.911s
```

Links
-----
* https://github.com/ManageIQ/azure-armrest/pull/316 : *Use respond_to? instead of methods.include?*
* https://github.com/ManageIQ/azure-armrest/pull/317 : *Lazy load json strings when possible*
* https://github.com/ManageIQ/azure-armrest/pull/318 : *Use `ArmrestCollection#next_link` in get_all_results*

### DIFF
--- a/lib/azure/armrest/storage_account_service.rb
+++ b/lib/azure/armrest/storage_account_service.rb
@@ -104,7 +104,7 @@ module Azure
       #
       # If you want a plain hash, use the list_account_keys method instead.
       #
-      def list_account_key_objects(account_name, group = configuration.resource_group)
+      def list_account_key_objects(account_name, group = configuration.resource_group, skip_accessors_definition = false)
         validate_resource_group(group)
 
         unless recent_api_version?
@@ -113,7 +113,7 @@ module Azure
 
         url = build_url(group, account_name, 'listKeys')
         response = rest_post(url)
-        JSON.parse(response.body)['keys'].map { |hash| StorageAccountKey.new(hash) }
+        JSON.parse(response.body)['keys'].map { |hash| StorageAccountKey.new(hash, skip_accessors_definition) }
       end
 
       alias list_storage_account_key_objects list_account_key_objects
@@ -172,7 +172,7 @@ module Azure
       # Note that for string values the comparison is caseless.
       #
       def list_all_private_images(filter = {})
-        storage_accounts = list_all(filter)
+        storage_accounts = list_all(filter.merge(:skip_accessors_definition => true))
         get_private_images(storage_accounts)
       end
 
@@ -187,7 +187,7 @@ module Azure
       #   sas.list_private_images(your_resource_group)
       #
       def list_private_images(group = configuration.resource_group)
-        storage_accounts = list(group)
+        storage_accounts = list(group, true)
         get_private_images(storage_accounts)
       end
 
@@ -273,30 +273,36 @@ module Azure
 
         Parallel.each(storage_accounts, :in_threads => configuration.max_threads) do |storage_account|
           begin
-            key = get_account_key(storage_account)
+            key = get_account_key(storage_account, true)
           rescue Azure::Armrest::ApiException
             next # Most likely due to incomplete or failed provisioning.
           else
             storage_account.access_key = key
           end
 
-          storage_account.containers.each do |container|
-            next if container.name =~ /^bootdiagnostics/i
-            storage_account.blobs(container.name).each do |blob|
-              next unless File.extname(blob.name).casecmp('.vhd').zero?
-              next unless blob.properties.lease_state.casecmp('available').zero?
+          init_opts = { :skip_accessors_definition => true }
+          storage_account.containers(storage_account.access_key, init_opts).each do |container|
+            next if container.name_from_hash =~ /^bootdiagnostics/i
+            storage_account.blobs(container.name_from_hash, storage_account.access_key, init_opts).each do |blob|
+              next unless File.extname(blob.name_from_hash).casecmp('.vhd').zero?
+              next unless blob.lease_state_from_hash.casecmp('available').zero?
 
               # In rare cases the endpoint will be unreachable. Warn and move on.
               begin
-                blob_properties = storage_account.blob_properties(blob.container, blob.name)
+                blob_properties = storage_account.blob_properties(
+                  blob[:container],
+                  blob.name_from_hash,
+                  storage_account.access_key,
+                  :skip_accessors_definition => true
+                )
               rescue Errno::ECONNREFUSED, Azure::Armrest::TimeoutException => err
                 msg = "Unable to collect blob properties for #{blob.name}/#{blob.container}: #{err}"
                 log('warn', msg)
                 next
               end
 
-              next unless blob_properties.respond_to?(:x_ms_meta_microsoftazurecompute_osstate)
-              next unless blob_properties.x_ms_meta_microsoftazurecompute_osstate.casecmp('generalized').zero?
+              next unless blob_properties[:x_ms_meta_microsoftazurecompute_osstate]
+              next unless blob_properties[:x_ms_meta_microsoftazurecompute_osstate].casecmp('generalized').zero?
 
               mutex.synchronize do
                 results << blob_to_private_image_object(storage_account, blob, blob_properties)
@@ -315,11 +321,11 @@ module Azure
         hash = blob.to_h.merge(
           :storage_account  => storage_account.to_h,
           :blob_properties  => blob_properties.to_h,
-          :operating_system => blob_properties.try(:x_ms_meta_microsoftazurecompute_ostype),
+          :operating_system => blob_properties[:x_ms_meta_microsoftazurecompute_ostype],
           :uri              => File.join(
-            storage_account.properties.primary_endpoints.blob,
-            blob.container,
-            blob.name
+            storage_account.blob_endpoint_from_hash,
+            blob[:container],
+            blob.name_from_hash
           )
         )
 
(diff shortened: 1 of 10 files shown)


NickLaMuro:
Think I have this where I am happy with it.  Taking off the `[WIP]`.

bzwei:
The ansible tower gem uses similar model to convert a hash to class. There however the methods are created per class rather than per instance. It still works through all hash keys in case there is any new key that was not discovered before. The good thing is it does not create new method if it has been added.

I haven't done any performance comparison but am thinking the other approach should help to boost the performance. I would not suggest `skip_accessors_definition` if the other approach performs close to this one.

NickLaMuro:
@bzwei Sorry for the delay on getting this... some quick background on the profiling I did do:

I originally did profiled a call of `.list_all_private_images` using to an azure provider, which I captured with VCR to allow quicker subsequent runs (the captured cassette was only 15 to 20 MB, so relatively small).  The account had about 200 storage accounts associated with this account, and I filtered by 'eastus', which brought it down to about 70.  Even with that, it required close to a gig to process (even with most of that garbage collected), and running it with the profiler enable took over 2 hours to process on my machine, and took it to a crawl in the process, but here is some of the important data from that report:

```
Total allocated: 4459823159 bytes (47621580 objects)
Total retained:  292214793 bytes (2755171 objects)

...

allocated objects by gem
  27079338  ruby-2.3.3/lib
  10929780  activesupport-5.0.6
   6248419  azure-armrest-0.8.2
   1878400  addressable-2.4.0
    623508  webmock-2.3.2
    355144  nokogiri-1.7.2
    190631  vcr-3.0.3
    171903  rest-client-2.0.2
     75270  azure-signature-0.2.3
     34654  json-2.0.4
     11940  netrc-0.11.0
     10149  http-cookie-1.0.3
      7401  domain_name-0.5.20170404
      2056  parallel-1.9.0
      1791  httpclient-2.8.3
      1196  other
```

For what it is worth, about all of the `ruby-2.3.3/lib` are from `rexml`, so basically `nokogiri`, and are probably produced within `libxml` and the associated c-ext, so those can be disregarded and/or are something we probably just have to live with, (that said, they are probably cleaned up manually outside of ruby's garbage collection mechanisms).  Also, this info is a bit out of date since some of this has been reduced with the 3 PRs mentioned in the description (see them for details).

* * *

So on to your points...

> The ansible tower gem uses similar model to convert a hash to class.  There however the methods are created per class rather than per instance.

Yup, and that is a big part of the hit.  After `rexml`, the part of the equation is then `activesupport`, and almost all of that is still in use (after excluding the fixes from the 3 PRs), is mostly caused by `.underscore` actually, when massaging the keys returned from azure to be more ruby-method-like.  As well as calling `instance_eval` after that part, so per object method instantiation is not good, and probably very repetitive on RAM/CPU (as you have mentioned).

> I haven't done any performance comparison but am thinking the other approach should help to boost the performance. I would not suggest skip_accessors_definition if the other approach performs close to this one.

No doubt this would be more performant to have these methods defined on class definition, but my major reason for not going that route is I was under the assumption that doing it this way was by design, so allow for "future proofing" changes in the API, while still allowing for a easy to use interface for the end user.

This approach effectively tries to maintain that assumption (however right or wrong it is), but only use it when it is actually needed by the end user.  If we decide that we would rather have a more "all knowing" client, then probably the ansible tower approach would be preferred, but that probably isn't a decision I am here to make ("I just work here...").

bronaghs:
Hi @NickLaMuro 
We saw the memory consumption increase incrementally when gathering private images, does this PR address that too?

NickLaMuro:
@bronaghs So without any proof, I would guess that "yes", it would use less memory incrementally since it doesn't need to build custom object definitions for each of the intermediate objects necessary to fetch private images (so storage accounts, storage account keys, blobs, properties, containers, etc.).  That said, as these fall out of scope, they should be garbage collected, so the memory should only grow if:

1.  A lot of these need to remain in memory for a period of time to get the next set
2.  Some other object down the chain references part or all of the object

I don't really solve for either of those with this, I just make it so that the amount of weight those intermediate objects hold is lessened.

That said, I don't really have the tooling in place when running these tests to measure this over time, so I don't have any evidence to prove this is the case besides what you can extrapolate from the memory samples taken at the end of the script.

bronaghs:
Can you use the same logging that @djberg96 used?, it printed out memory consumed every time a blob was read - so around [here](https://github.com/ManageIQ/azure-armrest/blob/master/lib/azure/armrest/storage_account_service.rb#L283). Then we could see if it grows.

bronaghs:
@NickLaMuro - Any reason why this PR would not not be backwards compatible with 5.8.1.5?

NickLaMuro:
@bronaghs Should work just fine as far as I know.  The changes here were done in such a way that only the intermediate models that `azure-armrest` is using to get the private images should be affected, and `manageiq` should be unaware of the change.  That said, I could be missing something, so it wouldn't hurt to have someone double check I did this right.

Regarding your other comment, I spoke with @djberg96 about getting that info, and have just been distracted with other items at the moment, but it is on my TODO for today to get some metrics about memory growth before and after this change.

bronaghs:
Sounds great - thank you @NickLaMuro

lib/azure/armrest/storage_account_service.rb
```diff
@@ -283,20 +283,26 @@ def get_private_images(storage_accounts)
...
             storage_account.blobs(container.name).each do |blob|
-              next unless File.extname(blob.name).casecmp('.vhd').zero?
-              next unless blob.properties.lease_state.casecmp('available').zero?
+              next unless File.extname(blob.name_from_hash).casecmp('.vhd').zero?
+              next unless blob.lease_state_from_hash.casecmp('available').zero?
 
               # In rare cases the endpoint will be unreachable. Warn and move on.
               begin
                 blob_properties = storage_account.blob_properties(blob.container, blob.name)
+                blob_properties = storage_account.blob_properties(
```
NickLaMuro: Oh man... I suck at rebasing... the line above this needs to be removed (since it is a dup).  Thankfully my benchmarking script caught this.

Will remove that shortly.

NickLaMuro:
Wanted to give a quick update:

First, didn't get to this as quickly as I was hoping today, so that did limit the amount of time that I was able to spend on this and get you the requested info.

Second, it looks like after the rebase of master and my updates from the conflicts created in #321 seem to have undone the savings that I showed previously.  Since I am still fixing a few things around that and investigating as to why this is the case, the metrics I promised have been delayed.

I might turn out that this now doesn't provide any improvements, though some slimmed down tests seem to show that they would, so I am looking into what I might have done wrong, and it is taking time.  Will try and get another update shortly tomorrow.

NickLaMuro:
Update:

So the fixup commits I added are all the things I did wrong when rebasing in the changes after https://github.com/ManageIQ/azure-armrest/pull/321 was merged.  I plan on rebasing those back into the previous commits, but haven't gotten around to doing so yet.

Those changes put the performance benchmarks back to what were shown in the description, though, they may be slightly changed because I did them without #321 in place, and I think I was also testing with a modified version of `activesupport` with the performance fix I did [here](https://github.com/rails/rails/pull/30782) in place.

I have also pushed a version of the changes from #324 with some updates to fix tests, address some review comments, and try to add some additional performance by combining the two efforts:

https://github.com/ManageIQ/azure-armrest/compare/master...NickLaMuro:light_base_model_mode_with_hash_model

I will try and report back later today with some benchmarks for both the above, and the differences with this PRs code and #324 on it's own.

bzwei:
@NickLaMuro are you done with this PR? Ready to be merged?

NickLaMuro:
@bzwei As far as code changes go, I think I am (the recent changes were only rebasing what I had as "FIXUP" commits into the last commit).

The only thing that was asked of me for this that remains outstanding is the comment from @bronaghs regarding getting memory overtime for this for the benchmarks I have.

I just recently put up the script that I had been using for benchmarking this whole time (after cleaning it up a bit and getting some sensitive data out of it) here:

https://gist.github.com/NickLaMuro/a1acfee84f8c58d8c55d72af6a4a1f9f

I can do the bit regarding getting memory over time, but as that script probably shows, I have not yet implemented doing that accurately yet (currently a WIP).  But if that is not really (or "no longer") necessary for increasing confidence before merging this, then no, I have nothing else to add at the moment.

bronaghs:
Hi @NickLaMuro - Benchmarking does not have to hold up the merging of this PR.

PR #320, opened 2017-09-26
Logger fix for StorageAccount

djberg96:
The `log` method isn't available to model objects, so we have to use the Configuration singleton method within the StorageAccount class.


djberg96:
The Hakiri warning is yet another nokogiri issue that is unrelated.

PR #321, opened 2017-09-27
Filter out diagnostic containers in get_private_images

djberg96:
This updates the `get_private_images` method (used by the `list_all_private_images` method) so that it filters out "diagnostic" containers before attempting to get blob information. These are containers that are automatically generated for VM's using unmanaged storage that turn on diagnostic information. They don't - or at least they shouldn't - store private images.

In practice this meant dispensing with the `all_blobs` method and handling the containers internally.

The net result was a slight increase in speed (since we're making fewer http requests), and about a 25 percent reduction in memory in informal testing (from 600mb to about 450mb). I cannot give a more precise amount since the amount of memory reduction will depend on just how many diagnostic containers a subscription has.

### DIFF
--- a/lib/azure/armrest/storage_account_service.rb
+++ b/lib/azure/armrest/storage_account_service.rb
@@ -276,26 +276,31 @@ module Azure
             key = get_account_key(storage_account)
           rescue Azure::Armrest::ApiException
             next # Most likely due to incomplete or failed provisioning.
+          else
+            storage_account.access_key = key
           end
 
-          storage_account.all_blobs(key, configuration.max_threads).each do |blob|
-            next unless File.extname(blob.name).casecmp('.vhd') == 0
-            next unless blob.properties.lease_state.casecmp('available') == 0
+          storage_account.containers.each do |container|
+            next if container.name =~ /^bootdiagnostics/i
+            storage_account.blobs(container.name).each do |blob|
+              next unless File.extname(blob.name).casecmp('.vhd').zero?
+              next unless blob.properties.lease_state.casecmp('available').zero?
 
-            # In rare cases the endpoint will be unreachable. Warn and move on.
-            begin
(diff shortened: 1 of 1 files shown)


lib/azure/armrest/storage_account_service.rb
```diff
@@ -276,26 +276,31 @@ def get_private_images(storage_accounts)
...
-              next
-            end
-
-            next unless blob_properties.respond_to?(:x_ms_meta_microsoftazurecompute_osstate)
-            next unless blob_properties.x_ms_meta_microsoftazurecompute_osstate.casecmp('generalized') == 0
-
-            mutex.synchronize do
-              results << blob_to_private_image_object(storage_account, blob, blob_properties)
+          storage_account.containers.each do |container|
+            next if container.name =~ /diagnostic/i
```
bronaghs: @djberg96 Who sets the name of the diagnostic container? Will it always contain the string "diagnostic"?

djberg96: @bronaghs Azure does. I guess the name is theoretically configurable after the fact, but by default it will contain the string "diagnostics" if, for example, boot diagnostics are enabled in the Azure portal.

bronaghs: OK thanks

bzwei: Should we skip if only a partial match? should it be `/^diagnostic$/i`?

djberg96: Yes. The bootdiagnostic containers, for example, are automatically appended with a mix of VM name and guid, e.g. "bootdiagnostics-somevm-xyz123abc".

That said, it looks like I could alter this from `/diagnostic/` to `/^bootdiagnostics/` as I don't see any other diagnostic container names.

djberg96:
@bzwei Ok, tightened up the regex a bit.

PR #322, opened 2017-09-27
Upgrade parallel gem to 1.12.x, use require_relative

djberg96:
Updates the parallel gem from 1.9 to 1.12, which includes a fix from none other than Joe R as well.

Also, I got rid of the `$LOAD_PATH` mangling, and just use `require_relative` instead.

### DIFF
--- a/azure-armrest.gemspec
+++ b/azure-armrest.gemspec
@@ -1,7 +1,5 @@
 # coding: utf-8
-lib = File.expand_path('../lib', __FILE__)
-$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
-require 'azure/armrest/version'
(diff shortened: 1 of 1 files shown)


bronaghs:
@djberg96 - What is the benefit to upgrading?

djberg96:
@bronaghs The short version is that it fixes some issues that we may or may not be affected by, but it's best to play it safe. For the long version, unfortunately the parallel gem doesn't have a history file, but here's what I could glean from the commit log:

* Uses Etc.nprocessors as a default value for certain methods
* Fixes a shadow variable warning
* Adds support for Ruby 2.4 and Travis (Joe R)
* Fixes an issue with zombie process in certain modes
* Plays nicer with the better_errors gem
* Adds the `.any?` and `.all?` methods

PR #324, opened 2017-09-29
[WIP] Use ansible client hash modeling

djberg96:
Based on comments from @bzwei at https://github.com/ManageIQ/azure-armrest/pull/319, I borrowed the HashModel class from the ansible tower client gem. I then subclassed our own BaseModel on that.

https://github.com/ansible/ansible_tower_client_ruby/blob/master/lib/ansible_tower_client/hash_model.rb

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -45,6 +45,11 @@ module Azure
         super
       end
 
+      def []=(key, value)
+        @data[key] = value
+        self.class.send(:convert_value, key, value, self.class)
+      end
+
       def hash_to_model(klass_name, hash)
         model_klass =
           if self.class.const_defined?(klass_name, false)
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -1,70 +1,15 @@
 require 'active_support/core_ext/string/inflections'
 require 'pp'
+require_relative 'hash_model'
 
 module Azure
   module Armrest
-    # Base class for JSON wrapper classes. Each Service class should have
-    # a corresponding class that wraps the JSON it collects, and each of
-    # them should subclass this base class.
-    class BaseModel
-      # Initially inherit the exclusion list from parent class or create an empty Set.
-      def self.excl_list
-        @excl_list ||= superclass.respond_to?(:excl_list, true) ? superclass.send(:excl_list) : Set.new
-      end
-
-      private_class_method :excl_list
-
-      # Merge the declared exclusive attributes to the existing list.
-      def self.attr_hash(*attrs)
-        @excl_list = excl_list | Set.new(attrs.map(&:to_s))
-      end
-
-      private_class_method :attr_hash
-
-      attr_hash :tags
-
+    class BaseModel < Azure::Armrest::HashModel
       attr_accessor :response_headers
       attr_accessor :response_code
 
-      # Constructs and returns a new JSON wrapper class. Pass in a plain
-      # JSON string and it will automatically give you accessor methods
-      # that make it behave like a typical Ruby object. You may also pass
-      # in a hash.
-      #
-      # Example:
-      #   class Person < Azure::ArmRest::BaseModel; end
-      #
-      #   json_string = '{"firstname":"jeff", "lastname":"durand",
-      #     "address": { "street":"22 charlotte rd", "zipcode":"01013"}
-      #   }'
-      #
-      #   # Or whatever your subclass happens to be.
-      #   person = Person.new(json_string)
-      #
-      #   # The JSON properties are now available as methods.
-      #   person.firstname        # => 'jeff'
-      #   person.address.zipcode  # => '01013'
-      #
-      #   # Or you can get back the original JSON if necessary.
-      #   person.to_json # => Returns original JSON
-      #
-      def initialize(json)
-        # Find the exclusion list for the model of next level (@embed_model)
-        # '#' is the separator between levels. Remove attributes
-        # before the first separator.
-        @child_excl_list = self.class.send(:excl_list).map do |e|
-          e.index('#') ? e[e.index('#') + 1..-1] : ''
-        end
-
-        if json.kind_of?(Hash)
-          @hash = json
-        else
-          @hash = JSON.parse(json)
-          @json = json
-        end
-
-        __setobj__(@hash.dup)
-      end
+      attr_writer :resource_group
+      attr_writer :subscription_id
 
       def resource_group
         @resource_group ||= id[/resourcegroups\/(.*?[^\/]+)?/i, 1] rescue nil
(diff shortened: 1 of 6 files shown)


lib/azure/armrest/model/hash_model.rb
```diff
@@ -0,0 +1,140 @@
...
+          attr_model(method)
+
+          return unless generate_writer?
+          class_eval { define_method("#{method}=") { |val| @data[key] = val } }
+        end
+      end ## self
+
+      def initialize(json_or_hash)
+        raw_hash = json_or_hash.kind_of?(Hash) ? json_or_hash : JSON.parse(json_or_hash)
+        @data = Hash[raw_hash.collect { |key, value| [key, self.class.send(:convert_value, key, value, self)] }]
```
NickLaMuro: Is there a reason we are going from a `Hash`, to an `Array`, and back again here?  Couldn't this just be simplified to:

```ruby
def initialize(json_or_hash)
  @data = json_or_hash.kind_of?(Hash) ? json_or_hash : JSON.parse(json_or_hash)
  @data.each {|key, value| @data[key] = self.class.send(:convert_value, key, value, self) }
end
```

NickLaMuro: Is there a reason we are going from a `Hash`, to an `Array`, and back again here?  Couldn't this just be simplified to:

```ruby
def initialize(json_or_hash)
  @data = json_or_hash.kind_of?(Hash) ? json_or_hash : JSON.parse(json_or_hash)
  @data.each {|key, value| @data[key] = self.class.send(:convert_value, key, value, self) }
end
```

lib/azure/armrest/model/hash_model.rb
```diff
@@ -0,0 +1,140 @@
...
+          # Merge the declared attributes to the modelized list
+          @modelized_list = modelized_list | Set.new(attrs.map(&:to_s))
+        end
+
+        # Return the list of modelized attributes
+        def modelized_list
+          @modelized_list ||= Set.new
+        end
+
+        def convert_value(key, value, parent)
```
NickLaMuro: So, this is still going to be costly to run, since there is a lot of `#underscore`, `#camelize`, and `#singularize` that will have to happen for each model instantiated, which was a lot of what I was trying to avoid in https://github.com/ManageIQ/azure-armrest/pull/319

Not sure of a good way around this though, without doing the approach in #319, and I fully admit that PR is kinda ugly and has it's own flaws as well.

lib/azure/armrest/model/hash_model.rb
```diff
@@ -0,0 +1,140 @@
...
+        # Default to false, to be overridden by subclasses.
+        #
+        def attr_excluded?(_attr_name)
+          false
+        end
+
+        # Declare attributes that need to be converted to Model
+        def attr_model(*attrs)
+          # Merge the declared attributes to the modelized list
+          @modelized_list = modelized_list | Set.new(attrs.map(&:to_s))
```
NickLaMuro: Hmmm... since this is a class method, is this thread safe?

lib/azure/armrest/model/hash_model.rb
```diff
@@ -0,0 +1,140 @@
...
+          method = key_to_attribute(key)
+          new_val =
+            if attr_excluded?(method)
+              value
+            else
+              # Must deal with nested models
+              case value
+              when Array
+                value.collect do |elem|
+                  elem.kind_of?(Hash) ? parent.send(:hash_to_model, method.camelize.singularize, elem) : elem
```
NickLaMuro: Would possibly a cached Hash for `key_to_attribute` and `method.camelize_sigularize` be helpful here?  I realize there would be thread safety issues with that as well that would need to be addressed, but it would avoid the costly (and repetitive) operations of doing these methods multiple times for every object instantiated.

lib/azure/armrest/model/hash_model.rb
djberg96: @NickLaMuro Updated.

lib/azure/armrest/model/hash_model.rb
djberg96: We do the similar stuff in the current code, do we not?

lib/azure/armrest/model/hash_model.rb
djberg96: I think the way around it is to instead to implement "as_hash" methods on ResourceGroupBasedService that users can use when they just want a plain hash and don't need intermediate objects.

So, you would have `get_as_hash`, `list_as_hash`, etc.

lib/azure/armrest/model/hash_model.rb
NickLaMuro: Not so much, but I will try an explain with an example.

Say we were making a call that made use of the `Parallel` gem to make multiple concurrent requests.  In the process of doing that, each of those are coming back at different times, instantiating there classes individually, and mutating `@modelized_list` at different points (with potentially different data).  This was not an issue with the previous implementation since any of the things being changed were on the instance level, and not shared memory of the class definition/class variables between multiple threads.

That said, with the `|` being done here, that might mitigate or even completely avoid issues because that would remove any duplicates being added.  This hurts my brain even to think about, let alone explain, so I could be over thinking this as well (part of the reason I brought this up as a question, not a statement 😉 )

lib/azure/armrest/model/hash_model.rb
NickLaMuro: Sorry, I was really more curious from a standpoint of what was being done here in the original code base, and not that it needed to be changed.  This might be better answered by someone (Brandon?) who actually wrote it.

lib/azure/armrest/model/hash_model.rb
NickLaMuro: Sure, I like that, or maybe just an option in those methods to avoid the object instantiation portion (since everything else should be the same).

NickLaMuro: And I guess not being familiar with the codebase myself, I didn't realize that the service models are the only things involved with making the requests to the API, where I was under the "Rails" mind set that the models would also be in charge of making calls.  That means doing what you are suggesting here really wouldn't be that involved.  Cool!

djberg96: To make things interesting, the StorageAccount model *does* make requests, but it's the only model that does. That, however, could change in the future.

NickLaMuro: Darn it!  I thought there was a model that did that!

Anyway, looks like that is an exception to the rule, and not the standard.

lib/azure/armrest/model/hash_model.rb
```diff
@@ -0,0 +1,138 @@
...
+              end
+            end
+
+          add_accessor_methods(method, key)
+          new_val
+        end
+
+        def add_accessor_methods(method, key)
+          method = "_#{method}" if instance_methods.include?(method.to_sym)
+          class_eval { define_method(method) { @data[key] } }
```
NickLaMuro: So, if I understand this correctly, won't every model end up getting the `_method` plus the original eventually, given enough objects have been instantiated?

The first time we go through this method, it would work as expected (assuming something like `#tags` or the like isn't already defined), and define the `"#{method}"`.  But the second time through, we would then end up defining the `"_#{method}"` because there is nothing stopping this from being redefined otherwise.

Seems like we need some kind of cache in place to check to see if we have defined methods using `add_accessor_methods` prior to doing it again because there is really nothing stoping use from entering this method again after we have run through it the first time with the same args.

lib/azure/armrest/model/base_model.rb
```diff
@@ -74,114 +43,34 @@ def subscription_id
...
-      end
-
-      def eql?(other)
-        return false unless other.kind_of?(BaseModel)
-        __getobj__.eql?(other.__getobj__)
-      end
+      def initialize(json_or_hash)
+        @child_excl_list = self.class.send(:excl_list).map do |e|
+          e.index('#') ? e[e.index('#') + 1..-1] : ''
+        end
```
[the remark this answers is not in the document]
NickLaMuro: @bzwei Could we change `attr_hash` to something like this?

```diff
 def self.attr_hash(*attrs)
   @excl_list = excl_list | Set.new(attrs.map(&:to_s))
+  @child_excl_list = self.class.send(:excl_list).map do |e|
+    e.index('#') ? e[e.index('#') + 1..-1] : ''
+  end
 end
```

And then change the portion of `#hash_to_model` to this:

```diff
-child_excl_list = @child_excl_list
+child_excl_list = self.class.child_excl_list
 self.class.const_set(klass_name, Class.new(self.class) { attr_hash(*child_excl_list) })
```

We do a class lookup in the `if` statement prior, so this should be just fine to do (if I am not mistaken).  Even if I am mistaken that that would be globally available (previously this was the case, but with `hash_to_model`, I no longer think it is), the `child_excl_list` should only need to updated when ever `attr_hash` is called, which is something that is hard coded into the codebase, and isn't something that is changed dynamically (only assigned to nested classes).

NickLaMuro:
:raising_hand_man: My fault 😄

lib/azure/armrest/model/hash_model.rb
djberg96: Oops, my bad. I was trying to solve the case issue.

djberg96:
Going to close this for now since we went with Nick's approach. I'll leave the branch in case we ever come back to it.

PR #325, opened 2017-10-02
Relax nokogiri gem version

NickLaMuro:
There is currently a CVE out for `libxml2`, that affects `nokogiri`:

https://access.redhat.com/security/cve/cve-2017-9050

Unfortunately, so projects that depend on `azure-armrest` expect that the version of nokogiri is within the `1.7.x` release.  To allow for that (as insecure as it is), but make sure we try to use the latest version of `nokogiri` when possible, relax the version requirement on this gem so that we are secure as we can be without breaking other projects.

That said, we probably should make nokogiri-1.8.1 a minimum requirement in the v1.0 release though.

### DIFF
--- a/azure-armrest.gemspec
+++ b/azure-armrest.gemspec
@@ -22,7 +22,7 @@ behind the scenes.
   spec.add_dependency 'memoist', '~> 0.15.0'
   spec.add_dependency 'azure-signature', '~> 0.2.3'
   spec.add_dependency 'activesupport', '>= 4.2.2'
-  spec.add_dependency 'nokogiri', '~> 1.7.2'
+  spec.add_dependency 'nokogiri', '~> 1.7'
   spec.add_dependency 'addressable', '~> 2.4.0'
   spec.add_dependency 'parallel', '~> 1.12.0'
 


NickLaMuro:
Hmmm... guess this doesn't make Hakiri much happier (actually worse...), but it does install `1.8.1` on the test suite, so to me that is an improvement I guess...

djberg96:
@NickLaMuro Let's just go straight to 1.8.x if we can.

NickLaMuro:
@djberg96 So, I was thinking about this and there were two things that I didn't really like about doing that:

1. It would break manageiq for the time period where we were waiting on a updated gem to be published, and undoing the undone-er PR of https://github.com/ManageIQ/manageiq-gems-pending/pull/279 (#wordSalad) would also need to merged within that period as well.
2.  I am unsure if this makes sense as a patch release, and we probably would want to do this as a 9.0 release if we do.  That something you are fine with doing?

Not saying we shouldn't go to `1.8.1`, but curious if it makes sense in this circumstance, and if we should rush to force users to use it.  Thoughts?

djberg96:
My thought is that we should switch to Oga so I don't have to worry about this ever again. I'm tired of these nokogiri security releases with major version bumps.

Anyway, I vote for doing whatever will alleviate the pain.

NickLaMuro:
@djberg96 I guess my follow up to that would be "what pain are we trying to alleviate, and is there more than one?"

The original intention of this was to get Hakiri to stop complaining on every PR, but this solution does not solve that... unfortunately.  It does allow the security issue be resolved as needed in other repos, however, which in my opinion is the bigger concern, and doing things this way allows us to get there without ruffling too many feathers if we just wanted to do a release with only this change.

That said, I think master is already a bit ahead of the most recent gem version, so forcing a `1.8.1` version of `nokogiri` and bumping this to `v0.9.0` would be pretty safe.  For us, specifically, `manageiq-providers-azure` is [locked to `< 0.9.0`](https://github.com/ManageIQ/manageiq-providers-azure/blob/9142a6b/manageiq-providers-azure.gemspec#L16) as well.

Again, fine with either, just don't want to ruffle feathers doing this improperly.

djberg96:
Alright, going to merge this for now. We'll bump to 0.9.0 in the next release, which will have major changes in the BaseModel one way or another.

NickLaMuro:
Cool, and sorry for the trouble.

Feel free to do either (bump and lockdown nokogiri or switch to Oga), and let me know if you cut a release so I can try getting `manageiq` up to date with `nokogiri`.

PR #327, opened 2017-10-06
Fasterer travis

Fryguy:
- Add [sudo:false](https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments) and [cache:bundler](https://docs.travis-ci.com/user/caching/#Bundler) for faster Travis runs. 
- Bump tested Rubies to the latest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ language: ruby
 sudo: false
 cache: bundler
 rvm:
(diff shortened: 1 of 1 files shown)


Fryguy:
This specific PR won't be faster, but once merged, then master will cache the bundled gems and subsequent runs will be much faster

PR #329, opened 2017-10-07
Handle possibility of dollar sign in key for model generation.

djberg96:
Apparently it's possible for the keys returned by the `TemplateDeploymentService#list_deployment_operations` to contain a key with a "$" in it. Since we have no idea where else this could potentially happen, we need to account for that possibility anywhere. So, it is now replaced with a "_" as well.

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1495318

I added some specs and fixed one unrelated spec description while I was in there.


djberg96:
@miq-bot add_label fine/yes

djberg96:
Oops, ignore that label attempt. I'll apply that to the azure provider once I release 0.9.1.

PR #330, opened 2017-10-09
Use Oga instead of Nokogiri

djberg96:
Due to a fairly steady stream of CVE's for nokogiri (usually because of underlying issues in libxml), along with issues sync'ing versions against the core ManageIQ repo, I think it's best if we switch off of Nokogiri to another parser.

I've chosen Oga because it's still relatively fast and doesn't have a libxml dependency. I'm not too worried about any speed issues, since the amount of xml parsing that we do overall is fairly minimal, limited to certain methods on `StorageAccount` objects.

For more information see https://github.com/YorickPeterse/oga

I've added some specs for the affected methods as well.

### DIFF
--- a/azure-armrest.gemspec
+++ b/azure-armrest.gemspec
@@ -22,7 +22,7 @@ behind the scenes.
   spec.add_dependency 'memoist', '~> 0.15.0'
   spec.add_dependency 'azure-signature', '~> 0.2.3'
   spec.add_dependency 'activesupport', '>= 4.2.2'
-  spec.add_dependency 'nokogiri', '>= 1.8.1', '~> 1.8'
+  spec.add_dependency 'oga', '~> 2.11'
   spec.add_dependency 'addressable', '~> 2.4.0'
   spec.add_dependency 'parallel', '~> 1.12.0'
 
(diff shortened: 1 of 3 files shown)


Fryguy:
FWIW, I am :-1: on this change.  I really don't see the relatively rare CVEs to nokogiri as a reason to abandon it...yeah, it's a little confusing to get ManageIQ updated, but it doesn't take *that* long, and this doesn't really remove the burden of updating nokogiri from the ManageIQ ecosystem anyway.

I would also like to see some memory comparison numbers.

Fryguy:
Forgot to add that oga itself has a c-extension component anyway, so it's not immune itself to CVEs.  In fact, it's more likely that security issues will be found with libxml2 because of it's widespread use over oga with it's narrow usage, and while that means more CVEs to deal with, it also means safer code.

jrafanie:
@djberg96 yeah, it feels like we're trading one headache for another.  If there were features of oga that didn't exist in nokogiri, maybe it would be worth it but it seems like we're just trying to avoid the CVEs found in the more popular library.  

If there's still compilation that's required, we'd have to document what packages are needed on the various OS platforms whereas nokogiri is pretty well known and established.

djberg96:
Closed in favor of https://github.com/ManageIQ/azure-armrest/pull/336

PR #334, opened 2017-10-11
Convert Nokogiri to Ox

djberg96:
Don't like Oga? Ok, let's switch to Ox instead then. Ox's main advantages are that:

1) It's not nokogiri
2) Doesn't have external C library dependencies. It uses system header files only.
3) Is faster than Nokogiri
4) It uses less memory than Nokogiri

Below is a benchmarking script that I used:

```
# ox_vs_noko.rb
require 'benchmark'
require 'nokogiri'
require 'ox'

xml = %(\xEF\xBB\xBF
  <?xml version=\"1.0\" encoding=\"utf-8\"?>
  <EnumerationResults ServiceEndpoint=\"https://foo.blob.core.windows.net/\" ContainerName=\"vhds\">
  <Blobs>
    <Blob>
      <Name>xyz.vhd</Name>
      <Properties>
        <Last-Modified>Thu, 10 Nov 2016 22:39:07 GMT</Last-Modified>
        <Etag>0x8D409BA6193F0C3</Etag>
        <Content-Length>31457280512</Content-Length>
        <Content-Type>application/octet-stream</Content-Type>
        <Content-Language />
        <Content-MD5>hKdOjkaup7sB/nzkWeuhWA==</Content-MD5>
        <x-ms-blob-sequence-number>1</x-ms-blob-sequence-number>
        <BlobType>PageBlob</BlobType>
        <LeaseStatus>unlocked</LeaseStatus>
        <LeaseState>available</LeaseState>
        <ServerEncrypted>false</ServerEncrypted>
      </Properties>
    </Blob>
    <Blob>
      <Name>foo</Name>
      <Properties>
        <Last-Modified>Thu, 06 Jul 2017 13:21:34 GMT</Last-Modified>
        <Etag>0x8D4C471ECC46568</Etag>
        <Content-Length>1024</Content-Length>
        <Content-Type>application/octet-stream</Content-Type>
        <Content-Language />
        <x-ms-blob-sequence-number>0</x-ms-blob-sequence-number>
        <BlobType>PageBlob</BlobType>
        <LeaseStatus>unlocked</LeaseStatus>
        <LeaseState>available</LeaseState>
        <ServerEncrypted>false</ServerEncrypted>
      </Properties>
    </Blob>
  </Blobs>
  <NextMarker />
  </EnumerationResults>
)

Benchmark.bm(35) do |x|
  x.report("Nokogiri::XML") do
    10000.times{ Nokogiri::XML(xml) }
  end

  x.report("Ox.parse") do
    10000.times{ Ox.parse(xml) }
  end

  x.report("Nokogiri#xpath") do
    doc = Nokogiri::XML(xml)
    100000.times{ doc.xpath("//Blobs/Blob") }
  end

  x.report("Ox#location") do
    doc = Ox.parse(xml)
    100000.times{ doc.locate('EnumerationResults/Blobs/Blob') }
  end
end
```

RESULTS:

```
                                          user     system      total        real
Nokogiri::XML                         0.610000   0.060000   0.670000 (  0.667711)
Ox.parse                              0.200000   0.010000   0.210000 (  0.206469)
Nokogiri#xpath                        2.310000   0.050000   2.360000 (  2.369448)
Ox#location                           0.440000   0.010000   0.450000 (  0.442009)
```

Quick update: this was updated so that it's being submitted against the 1.0.x branch instead of master.

### DIFF
--- a/azure-armrest.gemspec
+++ b/azure-armrest.gemspec
@@ -22,7 +22,7 @@ behind the scenes.
   spec.add_dependency 'memoist', '~> 0.15.0'
   spec.add_dependency 'azure-signature', '~> 0.2.3'
   spec.add_dependency 'activesupport', '>= 4.2.2'
-  spec.add_dependency 'nokogiri', '>= 1.8.1', '~> 1.8'
+  spec.add_dependency 'ox', '~> 2.8'
   spec.add_dependency 'addressable', '~> 2.4.0'
   spec.add_dependency 'parallel', '~> 1.12.0'
 
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -1,6 +1,6 @@
 require 'azure-signature'
 require 'active_support/core_ext/hash/conversions'
-require 'nokogiri'
+require 'ox'
 
 module Azure
   module Armrest
@@ -202,16 +202,13 @@ module Azure
 
         response = file_response(key, query, 'get', nil, share)
 
-        doc = Nokogiri::XML(response.body)
+        doc = Ox.parse(response.body)
         results = []
 
-        doc.xpath('//EnumerationResults/Entries').each do |element|
-          element.xpath('//Directory').each do |dir|
-            results << ShareDirectory.new(Hash.from_xml(dir.to_s)['Directory'])
-          end
-          element.xpath('//File').each do |file|
-            results << ShareFile.new(Hash.from_xml(file.to_s)['File'])
-          end
+        doc.EnumerationResults.Entries.each do |element|
+          hash = Hash.from_xml(Ox.to_xml(element))
+          results << ShareDirectory.new(hash['Directory']) if hash['Directory']
+          results << ShareFile.new(hash['File']) if hash['File']
         end
 
         results.concat(next_marker_results(doc, :files, key, options))
@@ -442,10 +439,12 @@ module Azure
 
         response = blob_response(key, query)
 
-        doc = Nokogiri::XML(response.body)
+        doc = Ox.parse(response.body)
+        results = []
 
-        results = doc.xpath('//Containers/Container').collect do |element|
-          Container.new(Hash.from_xml(element.to_s)['Container'], skip_defs)
+        doc.EnumerationResults.Containers.each do |container|
+          hash = Hash.from_xml(Ox.to_xml(container))
+          results << Container.new(hash['Container'], skip_defs)
         end
 
         results.concat(next_marker_results(doc, :containers, key, options))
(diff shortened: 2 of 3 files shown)


djberg96:
As for memory, I tried comparing two simple scripts that basically do the same thing. The sleep call was mainly so I could double check via the Activity Monitor.

```
# mem_check_nokogiri.rb
require 'sys-proctable'
require 'nokogiri'

1000.times{
  doc = Nokogiri::XML(xml)
  doc.xpath("//Blobs/Blob")
}

sleep 2
p Sys::ProcTable.ps(Process.pid).rss / 1024.0 / 1024.0
```

```
# mem_check_ox.rb
require 'ox'
require 'sys-proctable'

1000.times{
  doc = Ox.parse(xml)
  doc.locate('EnumerationResults/Blobs/Blob')
}

sleep 2
p Sys::ProcTable.ps(Process.pid).rss / 1024.0 / 1024.0
```

In my informal testing, the Nokogiri script used 30-34mb of memory.

The Ox script used 8mb.

Fryguy:
> Doesn't have external C library dependencies.

https://github.com/ohler55/ox/tree/master/ext/ox

$ gem install ox
Fetching: ox-2.8.0.gem (100%)
Building native extensions.  This could take a while...

---

Am I missing something?

Fryguy:
To clarify, I'm not against replacing LibXML/Nokogiri, especially since this uses less memory, which is cool, but if the reason is to avoid CVEs, my argument from the Oga thread about custom C-code being less secure than throughly used C-code still stands.

Fryguy:
@djberg96 Can you tweak your script so that it keeps the XMLs around (maybe in an Array)?  The way the script it written, it's possible a GC could kick in and clear out all of the memory, so it's not a fair comparison.

djberg96:
@Fryguy meaning, no libxml and libxslt. It's using its own C files, which only include your system header files.

Fryguy:
>  It's using its own C files

That's precisely the problem with Oga.  It's C-code, which, by its nature, is subject to a number of low-level attacks.  If the goal is to avoid CVEs, changing from one bit of C-code to another doesn't really change anything.  Oga used ragel so it's more of a parser generator as opposed to ox, which is all hand-written (they even [wrote base64](https://github.com/ohler55/ox/blob/master/ext/ox/base64.c)), but in either case, you have the possibility of unsafe code.

djberg96:
> If the goal is to avoid CVEs, changing from one bit of C-code to another doesn't really change anything.

This sort of assumes that code quality is the same across all C libraries, when it could just be shitty code. I've seen no CVE's for Ox. Not that there aren't any, but at least if there is one I only have to worry about Ox itself, not libxml and libxslt. If we can save memory and improve speed along the way, so much the better. :)

Plus, the goal isn't *just* to avoid CVE's, it's also to get away from versioning conflicts with the core repo that have caused more headaches than they're worth on multiple occasions for a very small part of the azure-armrest library that's going to be used less and less in the future as more and more users switch to managed storage.

djberg96:
Closing since we're dispensing with XML parsers altogether and just using `Hash#from_xml`.

PR #336, opened 2017-10-15
Remove nokogiri, just use activesupport

djberg96:
We're already using `Hash.from_xml` from the `activesupport` library. It turns out, that's really all we need. This eliminates the need to explicitly require nokogiri, or any xml parsing library for that matter, as a dependency.

### DIFF
--- a/lib/azure/armrest/model/storage_account.rb
+++ b/lib/azure/armrest/model/storage_account.rb
@@ -1,6 +1,5 @@
 require 'azure-signature'
 require 'active_support/core_ext/hash/conversions'
-require 'nokogiri'
 
 module Azure
   module Armrest
@@ -194,27 +193,43 @@ module Azure
       # * marker
       # * maxresults
       # * timeout
+      # * all
+      #
+      # If the :all option is set to true, then this method will collect all
+      # records. Otherwise, it is capped at 5000 records by Azure, or whatever
+      # you set :maxresults to. If you set both the :all option and the :maxresults
+      # option, then all records will be collected in :maxresults batches.
       #
       def files(share, key = access_key, options = {})
         raise ArgumentError, "No access key specified" unless key
 
-        query = {:restype => 'directory', :comp => 'list'}.merge(options).to_query
+        query = "comp=list&restype=directory"
 
-        response = file_response(key, query, 'get', nil, share)
+        skip_defs = options[:skip_accessors_definition]
 
-        doc = Nokogiri::XML(response.body)
-        results = []
-
-        doc.xpath('//EnumerationResults/Entries').each do |element|
-          element.xpath('//Directory').each do |dir|
-            results << ShareDirectory.new(Hash.from_xml(dir.to_s)['Directory'])
-          end
-          element.xpath('//File').each do |file|
-            results << ShareFile.new(Hash.from_xml(file.to_s)['File'])
+        options.each do |okey, ovalue|
+          unless okey == :skip_accessors_definition
+            query += "&#{okey}=#{[ovalue].flatten.join(',')}"
           end
         end
 
-        results.concat(next_marker_results(doc, :files, key, options))
+        response = file_response(key, query, 'get', nil, share)
+
+        hash = Hash.from_xml(response.body)['EnumerationResults']['Entries']
+        results = []
+
+        if hash && hash['Directory']
+          Array.wrap(hash['Directory']).each { |dir| results << ShareDirectory.new(dir, skip_defs) }
+        end
+
+        if hash && hash['File']
+          Array.wrap(hash['File']).each { |file| results << ShareFile.new(file, skip_defs) }
+        end
+
+        if options[:all] && hash['NextMarker']
+          options[:marker] = hash['NextMarker']
+          results.concat(files(share, key, options))
+        end
 
         results
       end
@@ -407,12 +422,18 @@ module Azure
       # * maxresults
       # * include
       # * timeout
+      # * all
       #
       # By default Azure uses a value of 5000 for :maxresults.
       #
       # If the :include option is specified, it should contain an array of
       # one element: metadata. More options may be added by Microsoft
       # at a later date.
+      #
+      # If the :all option is set to true, then this method will collect all
+      # records. Otherwise, it is capped at 5000 records by Azure, or whatever
+      # you set :maxresults to. If you set both the :all option and the :maxresults
+      # option, then all records will be collected in :maxresults batches.
       #
       # Example:
       #
@@ -423,10 +444,6 @@ module Azure
       #   p acct.containers(key)
       #   p acct.containers(key, :include => ['metadata'])
       #   p acct.containers(key, :maxresults => 1)
-      #
-      # In cases where a NextMarker element is found in the original response,
-      # another call will automatically be made with the marker value included
-      # in the URL so that you don't have to perform such a step manually.
       #
       def containers(key = access_key, options = {})
         raise ArgumentError, "No access key specified" unless key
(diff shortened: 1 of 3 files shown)


djberg96:
The failure is because `Hash#dig` isn't supported until 2.3. Can I drop support for 2.2?

bzwei:
@djberg96 How does `Hash.from_xml` perform?

djberg96:
@bzwei Not very well, but then it's hard to compare since `Nokogiri::XML(xml)` isn't returning a hash, but a document object, which we then have to iterate over via xpath, and then ultimately convert those results into a hash anyway, so I think the savings are somewhat illusory.

I haven't looked at the `ActiveSupport::XMLConverter` code to see how it's implemented.

djberg96:
@bzwei I should point out that my benchmarks were against 10000 iterations. If I just run a single `Hash.from_xml` call against an xml file with 1k file entries, for example, it's still pretty fast.

IMHO, it's not significant.

bzwei:
Sounds good to me 👍  @NickLaMuro do you have any inside of this replacement? 
We can take this change as long as we don't big negative impact.

Fryguy:
@jrafanie @chrisarcand I think so, but can we drop Ruby 2.2. support?

lib/azure/armrest/model/storage_account.rb
```diff
@@ -194,27 +193,38 @@ def directory_metadata(share, directory, key = access_key, options = {})
...
         raise ArgumentError, "No access key specified" unless key
 
-        query = {:restype => 'directory', :comp => 'list'}.merge(options).to_query
+        query = "comp=list&restype=directory"
+
+        skip_defs = options[:skip_accessors_definition]
+
+        options.each do |okey, ovalue|
+          unless okey == :skip_accessors_definition
+            query += "&#{okey}=#{[ovalue].flatten.join(',')}"
```
Fryguy: Prefer using URI.encode_www_form helper method to ensure things are escaped properly.

```ruby
URI.encode_www_form(okey => ovalue)
```

You could also just build up a big hash and pass it once...

```ruby
query_options = {:restype => 'directory', :comp => 'list'}.merge(options.except(:skip_accessors_definition))
query = URI.encode_www_form(query_options)
```

jrafanie:
> @jrafanie @chrisarcand I think so, but can we drop Ruby 2.2. support?

I wouldn't drop support for ruby 2.2 to use Hash#dig in a library.  I'd just write the 2.2 compatible code.

lib/azure/armrest/model/storage_account.rb
djberg96: No can do boss man, the `marker` can't be encoded or it blows up.

djberg96:
@jrafanie But....I'm LAZY.

jrafanie:
> @jrafanie But....I'm LAZY.

@djberg96 I can 'dig' that but think of the problems with having to bump the minor version

NickLaMuro:
> > @jrafanie But....I'm LAZY.
>
> @djberg96 I can 'dig' that but think of the problems with having to bump the minor version

(－‸ლ)

NickLaMuro:
@djberg96 @bzwei Here are some before and after metrics:


`0.9.1` to `djberg96:maxresults`

```
Before:                       After:

Mem: 429.796875MiB            Mem: 422.65234375MiB
                              
real    0m37.264s             real    0m26.880s
user    0m28.117s             user    0m21.666s
sys     0m0.725s              sys     0m1.019s
                              
                              
Mem: 421.98828125MiB          Mem: 409.1953125MiB
                              
real    0m39.103s             real    0m26.053s
user    0m33.773s             user    0m21.309s
sys     0m1.064s              sys     0m0.981s
                              
                              
Mem: 424.2890625MiB           Mem: 418.53515625MiB
                              
real    0m37.241s             real    0m26.368s
user    0m32.309s             user    0m21.493s
sys     0m0.867s              sys     0m1.007s
                              
                              
Mem: 428.91796875MiB          Mem: 443.0234375MiB
                              
real    0m36.892s             real    0m26.338s
user    0m32.550s             user    0m21.577s
sys     0m0.919s              sys     0m1.001s
```

This is what is currently in the branch, but it looks like there is a ~10sec speed up here.  👍

bzwei:
@NickLaMuro Great! Thanks!

chrisarcand:
> I wouldn't drop support for ruby 2.2 to use Hash#dig in a library. I'd just write the 2.2 compatible code.

Agreed; Probably shouldn't have dropped a still-supported version of Ruby in a library. Maintenance of Ruby 2.2 doesn't end until March 2018.

djberg96:
@chrisarcand @jrafanie Ok, I've submitted this against 1.0.x instead of master. So, we can set March 2018 date as our tentative target.

NickLaMuro:
@djberg96 a patch you could consider if you want this in sooner:

```diff
diff --git a/lib/azure/armrest.rb b/lib/azure/armrest.rb
index 0e62e86..b95288e 100644
--- a/lib/azure/armrest.rb
+++ b/lib/azure/armrest.rb
@@ -5,6 +5,7 @@ require 'addressable'
 require 'parallel'
 require 'memoist'
 require 'active_support/core_ext/hash/keys'
+require 'azure/armrest/core_ext/hash/dig'
 
 # The Azure module serves as a namespace.
 module Azure
diff --git a/lib/azure/armrest/core_ext/hash/dig.rb b/lib/azure/armrest/core_ext/hash/dig.rb
new file mode 100644
index 0000000..12f21b0
--- /dev/null
+++ b/lib/azure/armrest/core_ext/hash/dig.rb
@@ -0,0 +1,10 @@
+class Hash
+  def dig(*keys)
+    result   = self
+    last_key = keys.pop
+    keys.each do |key|
+      result = result[key] || {}
+    end
+    result[last_key]
+  end unless {}.respond_to?(:dig)
+end
```

Pre ruby-2.2, this is not memory efficient at all (adds about 50MB to the process in my tests above... though this is better than my first pass at doing this, which was about twice that), but it should allow you to maintain < `ruby-2.2.0` support while still keeping the implementation you have.

Something to consider.

chrisarcand:
> Ok, I've submitted this against 1.0.x instead of master. So, we can set March 2018 date as our tentative target.

I'm...confused as to how this gem is versioned and what that means. What's the difference between `master` and the `1.0.x` branch and why does the `1.0.x` still have a `0.9.x` version file in it? Didn't you already merge dropping Ruby 2.2. in `master`? ¯\\\(°_o)/¯ Might not be worth the time explaining to me I guess, I was just here to answer a question 😄

djberg96:
@NickLaMuro I was just being lazy, so I reworked it to just do existence checks. @jrafanie Ok, I stopped being lazy.

@chrisarcand No worries, it's pointing back at master now.

lib/azure/armrest/model/storage_account.rb
```diff
@@ -194,27 +193,43 @@ def directory_metadata(share, directory, key = access_key, options = {})
...
+          end
+        end
 
         response = file_response(key, query, 'get', nil, share)
 
-        doc = Nokogiri::XML(response.body)
+        hash = Hash.from_xml(response.body)['EnumerationResults']['Entries']
```
bzwei: Can it ever fail? Does `['EnumerationResults']` alway present?

lib/azure/armrest/model/storage_account.rb
```diff
@@ -194,27 +193,43 @@ def directory_metadata(share, directory, key = access_key, options = {})
...
 
-        doc.xpath('//EnumerationResults/Entries').each do |element|
-          element.xpath('//Directory').each do |dir|
-            results << ShareDirectory.new(Hash.from_xml(dir.to_s)['Directory'])
-          end
-          element.xpath('//File').each do |file|
-            results << ShareFile.new(Hash.from_xml(file.to_s)['File'])
-          end
+        if hash && hash['Directory']
+          Array.wrap(hash['Directory']).each { |dir| results << ShareDirectory.new(dir, skip_defs) }
```
bzwei: Does `hash` always contains `Directory` and `File`?

djberg96: No, but the Array.wrap will convert nil to `[]` if it's non-existent.

lib/azure/armrest/model/storage_account.rb
djberg96: In all my tests, yes, it's always present.

PR #340, opened 2017-10-24
Ensure value passed to sleep method is an integer

djberg96:
The `err.response.headers[:retry_after]` value is a stringified number instead of a plain number, so we must explicitly call `.to_i` on the value to `sleep` or an error will occur.

Note that we already do this for the `wait` method.

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -216,7 +216,9 @@ module Azure
             if tries <= max_retries
               msg = "A rate limit or server side issue has occurred [#{err.http_code}]. Retry number #{tries}."
               Azure::Armrest::Configuration.log.try(:log, Logger::WARN, msg)
-              sleep_time = err.response.headers[:retry_after] || 30
+              sleep_time = (err.response.headers[:retry_after] || 30).to_i
+              sleep_time = 5 if sleep_time < 5     # 5 second minimum
+              sleep_time = 120 if sleep_time > 120 # 2 minute maximum
               sleep(sleep_time)
               retry
             end


lib/azure/armrest/armrest_service.rb
```diff
@@ -217,7 +217,7 @@ def rest_execute(options, http_method = :get, encode = true, max_retries = 3)
               msg = "A rate limit or server side issue has occurred [#{err.http_code}]. Retry number #{tries}."
               Azure::Armrest::Configuration.log.try(:log, Logger::WARN, msg)
               sleep_time = err.response.headers[:retry_after] || 30
-              sleep(sleep_time)
+              sleep(sleep_time.to_i)
```
blomquisg: If `:retry_after` is a non integer, this will result in `0` for `sleep_time`.  Is that desired?  Or, should it resort to `30`?

I'm worried that `sleep_time == 0` will result in trying a number of requests over and over without any sleep, and never allowing the API rate limit to reset.

djberg96: @blomquisg as far as I know `:retry_after` will always be a stringified number, which `.to_i` converts to the equivalent integer, e.g. `"25".to_i` is 25, not 0.

blomquisg: K.  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After says that it *could* be a date or a number of seconds.

I'm not sure if Azure has some inconsistency in their value.  If they are sometimes passing back a date, this could be bad for us.  The only way I can see to monitor that is to log out what it's saying for the requested timeout duration.  That might be a good idea anyway in the warn line above this.

djberg96: I've never seen a date passed back for that value, and all of their REST API documentation that I've come across says it will be number of seconds, e.g. https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-request-limits.

blomquisg: The thing that's really concerning to me is that this fails today only *sometimes*.  That means that sometimes this current code works. 

That means one of a few things:

The response header value of `retry_after`
a. sometimes comes back as an int, and sometimes comes back as a string, but it's always an integer value (25, or "25")
b. sometimes comes back as empty string, and sometimes comes back as an int ("", or 25)
c. sometimes comes back as a non-int, and sometimes comes back as an int ("Wed, 21 Oct 2015 07:28:00 GMT", or 25)

In any case, I think this should definitely handle the case where the resulting `sleep_time` is `0`.  That way you don't have to care about the special cases for the value in the `retry_after`.

djberg96:
@blomquisg Ok, I've ensured a 5-second minimum and a 2-minute maximum. :)

lib/azure/armrest/armrest_service.rb
```diff
@@ -216,7 +216,9 @@ def rest_execute(options, http_method = :get, encode = true, max_retries = 3)
             if tries <= max_retries
               msg = "A rate limit or server side issue has occurred [#{err.http_code}]. Retry number #{tries}."
               Azure::Armrest::Configuration.log.try(:log, Logger::WARN, msg)
-              sleep_time = err.response.headers[:retry_after] || 30
+              sleep_time = (err.response.headers[:retry_after] || 30).to_i
+              sleep_time = 5 if sleep_time < 5     # 5 second minimum
+              sleep_time = 120 if sleep_time > 120 # 2 minute maximum
```
Fryguy: Too bad we can't use the new #clamp method yet

https://blog.bigbinary.com/2016/12/13/ruby-2-4-adds-comparable-clamp-method.html

Fryguy: If you're using more_core_extensions, we do have clamp, but it doesn't look like this gem uses more_core_extensions

https://github.com/ManageIQ/more_core_extensions/blob/master/lib/more_core_extensions/core_ext/numeric/clamp.rb

blomquisg:
An example of the traceback that we noticed in the logs.

```
[----] E, [2017-10-20T06:18:39.796012 #3636:120f138] ERROR -- : [TypeError]: can't convert String into time interval  Method:[rescue in perf_collect_metrics]
[----] E, [2017-10-20T06:18:39.796239 #3636:120f138] ERROR -- : /opt/rh/cfme-gemset/gems/azure-armrest-0.8.5/lib/azure/armrest/armrest_service.rb:220:in `sleep'
/opt/rh/cfme-gemset/gems/azure-armrest-0.8.5/lib/azure/armrest/armrest_service.rb:220:in `rescue in rest_execute'
/opt/rh/cfme-gemset/gems/azure-armrest-0.8.5/lib/azure/armrest/armrest_service.rb:209:in `rest_execute'
/opt/rh/cfme-gemset/gems/azure-armrest-0.8.5/lib/azure/armrest/armrest_service.rb:298:in `rest_execute'
/opt/rh/cfme-gemset/gems/azure-armrest-0.8.5/lib/azure/armrest/armrest_service.rb:314:in `rest_post'
/opt/rh/cfme-gemset/gems/azure-armrest-0.8.5/lib/azure/armrest/storage_account_service.rb:93:in `list_account_keys'
/opt/rh/cfme-gemset/bundler/gems/manageiq-providers-azure-ae2e0e06ee85/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb:183:in `raw_metrics_for_counter'
/opt/rh/cfme-gemset/bundler/gems/manageiq-providers-azure-ae2e0e06ee85/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb:159:in `block (2 levels) in metrics_by_counter_name'
/opt/rh/cfme-gemset/bundler/gems/manageiq-gems-pending-8f7d63f441d5/lib/gems/pending/util/extensions/miq-benchmark.rb:11:in `realtime_store'
/opt/rh/cfme-gemset/bundler/gems/manageiq-gems-pending-8f7d63f441d5/lib/gems/pending/util/extensions/miq-benchmark.rb:30:in `realtime_block'
/opt/rh/cfme-gemset/bundler/gems/manageiq-providers-azure-ae2e0e06ee85/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb:158:in `block in metrics_by_counter_name'
/opt/rh/cfme-gemset/bundler/gems/manageiq-providers-azure-ae2e0e06ee85/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb:155:in `each'
/opt/rh/cfme-gemset/bundler/gems/manageiq-providers-azure-ae2e0e06ee85/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb:155:in `each_with_object'
/opt/rh/cfme-gemset/bundler/gems/manageiq-providers-azure-ae2e0e06ee85/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb:155:in `metrics_by_counter_name'
/opt/rh/cfme-gemset/bundler/gems/manageiq-providers-azure-ae2e0e06ee85/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb:126:in `perf_capture_data_azure'
/opt/rh/cfme-gemset/bundler/gems/manageiq-providers-azure-ae2e0e06ee85/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb:98:in `block (2 levels) in perf_collect_metrics'
/opt/rh/cfme-gemset/bundler/gems/manageiq-providers-azure-ae2e0e06ee85/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb:115:in `with_metrics_services'
/opt/rh/cfme-gemset/bundler/gems/manageiq-providers-azure-ae2e0e06ee85/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb:97:in `block in perf_collect_metrics'
/var/www/miq/vmdb/app/models/ext_management_system.rb:378:in `with_provider_connection'
/opt/rh/cfme-gemset/bundler/gems/manageiq-providers-azure-ae2e0e06ee85/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb:96:in `perf_collect_metrics'
/var/www/miq/vmdb/app/models/metric/ci_mixin/capture.rb:6:in `perf_collect_metrics'
```

lib/azure/armrest/armrest_service.rb
djberg96: @Fryguy didn't know about that, could be useful in the future!

PR #343, opened 2017-10-25
[FINE] Changes for the 0.8.7 release

bronaghs:
Updates the CHANGES file with the change included in 0.8.7

I noticed 0.8.6 is not included in this file, but I did not want to add an entry because I do not see the [PR](https://github.com/ManageIQ/azure-armrest/pull/329) in the 0.8.x commit log
@djberg96 - can you investigate?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,6 @@
+= 0.8.7 - 25-Oct-2017
+* Ensure value passed to sleep method is an integer.
+
 = 0.8.6 - 23-Oct-2017
 * Backported the '$' model key fix from 0.9.1.
 


djberg96:
Apparently I made the change locally, but never pushed it up. Not a huge deal. Just add this to the CHANGES file:

```
= 0.8.6 - 23-Oct-2017
* Backported the '$' model key fix from 0.9.1.
```

bronaghs:
Thanks @djberg96 - Updated PR with the 0.8.6 entry added

bronaghs:
Actually @djberg96 (When you are not on PTO) - I didn't see the PR in the commit log for 0.8.x, do you know why?

djberg96:
Ugh, it didn't get pushed up. The `rake release` task blows up when it's not master, and I guess I never did it manually.

bronaghs:
What is confusing is that the fix is in the 0.8.6 files eg https://github.com/ManageIQ/azure-armrest/blob/v0.8.6/spec/models/base_model_spec.rb#L217..#L223
but the commit isnt in the log...

djberg96:
Just a quick followup. For backports I've just been doing cherry-picks and pushing them up. I made the changes locally for 0.8.6 but never pushed up the commits. The 0.8.6 gem DOES have the expected fixes, though. It was only the repo that got out of sync.

Anyway, the commits were added back in and pushed up, so a small bit of confusion as we reworked this PR.

djberg96:
LGTM 👍

Issue #332, opened 2017-10-11
Some current test failures on Windows


djberg96:
Windows 8
ruby 2.4.2p198 (2017-09-14 revision 59899) [i386-mingw32]

While not a high priority, this library should be platform neutral, so we shouldn't be seeing failures on any platform. I suspect this is just the way we've written our tests rather than a core library issue, but thought I should post.

```
Failures:

  1) Azure::Armrest::Configuration instances tokens token generation caches the token to be reused for the same cl
     Failure/Error: expect(subject.token).to eql(token)

       expected: "Bearer eyJ0eXAiOiJKV1Q"
            got: "xxx"

       (compared using eql?)
     # ./spec/configuration_spec.rb:185:in `block (5 levels) in <top (required)>'

  2) BaseModel inspection methods defines a pretty_print method when pp is available
     Failure/Error: expect(base.pretty_inspect).to match(expected)

       expected "#<Azure::Armrest::BaseModel:0x03b95f98 name=\"test\", array=[\"stuff\"], age=33>\n" to match /\A#
::Armrest::BaseModel:0x\h+\n/
       Diff:
       @@ -1,2 +1,2 @@
       -/\A#<Azure::Armrest::BaseModel:0x\h+\n/
       +#<Azure::Armrest::BaseModel:0x03b95f98 name="test", array=["stuff"], age=33>

     # ./spec/models/base_model_spec.rb:146:in `block (3 levels) in <top (required)>'

  3) BaseModel #attr_from_hash maps it source location to the correct file
     Failure/Error: expect(subject.method(:first_name_from_hash).source_location).to include(__FILE__)
       expected ["C", 0] to include "C:/Users/Daniel/Dev/azure-armrest-djberg96/spec/models/base_model_spec.rb"
     # ./spec/models/base_model_spec.rb:291:in `block (3 levels) in <top (required)>'
     # ./spec/models/base_model_spec.rb:279:in `block (3 levels) in <top (required)>'

  4) BaseModel #attr_from_hash with multiple attributes maps it source location of each method to the correct file
     Failure/Error: expect(subject.method(:first_name_from_hash).source_location).to include(__FILE__)
       expected ["C", 0] to include "C:/Users/Daniel/Dev/azure-armrest-djberg96/spec/models/base_model_spec.rb"
     # ./spec/models/base_model_spec.rb:313:in `block (4 levels) in <top (required)>'
[...]

PR #344, opened 2017-10-25
Use caller_locations instead of caller

NickLaMuro:
Ruby already has a mechanism in place for getting the callstack path locations and line numbers that doesn't require string splitting/parsing.

This also fixes a bug on Windows systems when trying to split callstacks with file paths in them, since most windows file paths include something like 'C:/path/to/file', which has an additional ':' that is not expected by the codebase.

Fixes some of https://github.com/ManageIQ/azure-armrest/issues/332

cc @djberg96


djberg96:
All tests pass on my Windows 10 box.

PR #345, opened 2017-10-26
Fix StorageAccountService#list_private_images

NickLaMuro:
When speeding up the `list_all_private_images` method, the `list_all` method was updated with the `skip_accessors_definition` boolean flag to that and associated methods.  This was not, though, applied to the `StorageAccountService#list_private_images` method, which works on just a single storage account instead of all of them.

This updates `#list` in both the `StorageAccountService` version, and the base class to allow it to support this flag.

Links
-----
* https://bugzilla.redhat.com/show_bug.cgi?id=1506667
* Introduced in https://github.com/ManageIQ/azure-armrest/pull/319

### DIFF
--- a/lib/azure/armrest/resource_group_based_service.rb
+++ b/lib/azure/armrest/resource_group_based_service.rb
@@ -72,14 +72,14 @@ module Azure
       # Returns an ArmrestCollection, with the response headers set
       # for the operation as a whole.
       #
-      def list(rgroup = configuration.resource_group)
+      def list(rgroup = configuration.resource_group, skip_accessors_definition = false)
         validate_resource_group(rgroup)
 
         url = build_url(rgroup)
         url = yield(url) || url if block_given?
         response = rest_get(url)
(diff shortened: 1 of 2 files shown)


bronaghs:
@djberg96 - Can you test this by provisioning an instance? also, can we add some specs?

NickLaMuro:
> also, can we add some specs?

So this is what I was hoping to do as well, but wasn't sure how we wanted to approach it.  Right now, I don't think there are any tests in place that attempt to make API calls, and stubbing that kinda of stuff in my experience leads to the API changing and then tests giving false positives of passing (that said, probably better than the issue we ran into here).

Anyway, didn't want to make a decision on how we wanted to go about that in this PR, so I didn't.  That said, wouldn't mind opening up another issue to discuss how we would do testing for this in the future.

djberg96:
I will add some specs in a separate PR as the code is definitely broken currently.

PR #346, opened 2017-10-30
Added some specs for list and list_all methods

djberg96:
This is meant as a followup to https://github.com/ManageIQ/azure-armrest/pull/345.

This adds specs for the list and list_all methods for both StorageAccountService and VirtualMachineService. It includes specs for skipping the accessor definitions as well.

### DIFF
--- a/spec/fixtures/unmanaged_storage_accounts.json
+++ b/spec/fixtures/unmanaged_storage_accounts.json
@@ -1,0 +1,99 @@
+{"value":
+  [
+    {
+      "accessTier": null,
+      "creationTime": "2016-11-10T21:20:55.058463+00:00",
+      "customDomain": null,
+      "encryption": null,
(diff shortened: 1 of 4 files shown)


NickLaMuro:
@djberg96 Thanks for doing this, this will be really helpful in the future I am sure.

Curious if you want to document somewhere the process of generating the json files you used as fixtures, in case we have to redo it in the future.  I figured steps should include what was provisioned to get the data, what was replaced with dummy data (to prevent sensitive data from being stored in these files), etc.

djberg96:
@NickLaMuro I just used the Azure CLI and, in cases where there were too many results, I cherry picked a few out of the output. The only thing that really needs replacing for security reasons is the subscription ID, though I also renamed the resource group for testing purposes.

I can document this in the README I suppose.

NickLaMuro:
@djberg96 fair enough.  Though, if you don't want to clutter your main `README.md`, you probably could just toss a separate `README.md` into the `spec/fixtures/` dir so it is gear specifically for creating new fixtures.

PR #347, opened 2017-11-01
Don't append api-version string if already included

djberg96:
When the `build_id_string` method was refactored, I accidentally introduced a bug in the `TemplateDeploymentService#delete_associated_resources` method. That method uses a combination of `build_id_string` and `delete_by_id`. The problem is that both methods append an `api-version` string, resulting in an invalid URL.

This PR modifies the `get_by_id` and `delete_by_id` methods so that the api-version string is only appended if it's not already present.

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1497175

### DIFF
--- a/lib/azure/armrest/resource_group_based_service.rb
+++ b/lib/azure/armrest/resource_group_based_service.rb
@@ -132,10 +132,9 @@ module Azure
       #
       def get_by_id(id_string)
         info = parse_id_string(id_string)
-        api_version = api_version_lookup(info['provider'], info['service_name'], info['subservice_name'])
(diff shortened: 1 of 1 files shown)


bzwei:
Looks like we can consolidate the changes into one common method like `id_string_to_url`

bzwei:
@djberg96 can you add link https://bugzilla.redhat.com/show_bug.cgi?id=1497175 to the commit comment so that the BZ gets the notification?

djberg96:
@bzwei Ok, refactored into a common method.

PR #349, opened 2017-11-21
Add support for HDInsight services

01100010011001010110010101110000:
This adds support for HDInsight cluster service and its application subservice.


lib/azure/armrest/hdinsight/hdinsight_application_service.rb
```diff
@@ -0,0 +1,11 @@
+module Azure
+  module Armrest
+    module HDInsight
+      class HDInsightApplicationService < ResourceGroupBasedSubservice
+        def initialize(armrest_configuration, options = {})
+          super(armrest_configuration, 'clusters', 'applications', 'Microsoft.HDInsight', options)
```
bzwei: `super` only accepts 4 arguments

01100010011001010110010101110000: I think `ResourceGroupBasedSubService#initialize` takes five arguments, and that would mean a super call for that method would as well, yes? None of the arguments for its initalize method have defaults or anything

bzwei: My bad 👎 .  I did not notice this is a sub service.

01100010011001010110010101110000: No worries, glad I didn't post something stupid for all the world to see 👍

spec/hdinsight_application_service_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  context "inheritance" do
+    it "is a subclass of ResourceGroupBasedSubservice" do
+      ancestors = Azure::Armrest::HDInsight::HDInsightApplicationService.ancestors
+      expect(ancestors).to include(Azure::Armrest::ResourceGroupBasedSubservice)
+    end
+  end
+
+  context "constructor" do
+    it "returns a Sql::SqlDatabaseService instance as expected" do
```
bzwei: A typo here. Evidently you copied and pasted an existing spec. Can you change the name `sds` to something more meaning to this test, such as `hd_ias`?

spec/hdinsight_cluster_service_spec.rb
```diff
@@ -0,0 +1,63 @@
+#############################################################################
+# hdinsight_cluster_service_spec.rb
+#
+# Test suite for the Azure::Armrest::HDInsight::HDInsightClusterService class.
+#############################################################################
+require 'spec_helper'
+
+describe "HDInsight::HDInsightClusterService" do
+  before { setup_params }
+  let(:server) { Azure::Armrest::HDInsight::HDInsightClusterService.new(@conf) }
```
bzwei: `server` or `service`?

spec/hdinsight_cluster_service_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+  end
+
+  context "constructor" do
+    it "returns a HDInsightCluster instance as expected" do
+      expect(server).to be_kind_of(Azure::Armrest::HDInsight::HDInsightClusterService)
+    end
+  end
+
+  context "accessors" do
```
bzwei: I am thinking we only need the first two tests. All the rest tests are from base class, which are already tested. You can delete them all. Same for the other test.

01100010011001010110010101110000: I don't see a spec for the subservice class, are those tests all done in the base service class? Also, yes, I shamelessly ripped off the SQL database and SQL server specs

PR #354, opened 2018-01-18
Use name_from_hash in exception handlers

djberg96:
When we switched over to the `skip_accessors` approach, we missed a couple places where the `.name` property isn't available. Instead, we must use `name_from_hash`.


djberg96:
@NickLaMuro Look good?

PR #355, opened 2018-01-19
Managed Disk Read Semantics Change

jerryk55:
Azure Managed Disk Instances have been taking exceptionally long in
some locations.  It turns out that the semantic that azure-armrest
uses to perform reads via the get_blob_raw method is the culprit.
The method gets a SAS URL - an access token to the disk - each
time it performs a read operation.  The call to get the access token
is asynchronous and if it does not return immediately with success when
poll is called the gem sleeps for 5 seconds.  This adds 5 seconds to
*every* call to get data.

get_blob_raw is being left in the gem for compatibility purposes.
However the new semantic is this:

Call open on the service (disk service or snapshot service).
Open returns a ManagedDisk object.
Call read on the ManagedDisk object as many times as you need, then call
close on the ManagedDisk object when finished.

Under the covers the ManagedDisk will contain the SAS URL to be used.  It should
only be obtained once unless it times out, in which case the caller is expected
to get a new ManagedDisk object to continue.

https://github.com/ManageIQ/manageiq-smartstate/pull/50 is also required for this fix to go forward and it depends on this PR.  It will have to be backported
to Gaprindashvilli and in the manageiq-gems-pending repo to the Fine release.  

@blomquisg @djberg96 @bronaghs @roliveri for your reading and reviewing pleasure.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1508154

### DIFF
--- a/lib/azure/armrest/storage/managed_disk.rb
+++ b/lib/azure/armrest/storage/managed_disk.rb
@@ -22,8 +22,8 @@ module Azure
             @storage_service.close(@disk_name, @resource_group)
             @sas_url = nil
           end
-        end # ManagedDisk
-      end # ManagedStorageHelper
-    end # Storage
-  end # Armrest
-end # Azure
+        end
+      end
+    end
+  end
+end
--- a/lib/azure/armrest/storage/managed_disk.rb
+++ b/lib/azure/armrest/storage/managed_disk.rb
@@ -15,7 +15,21 @@ module Azure
           end
 
           def read(options = {})
-            @storage_service.read(@sas_url, options)
+            retries = 0
+            begin
+              @storage_service.read(@sas_url, options)
+            rescue Azure::Armrest::ForbiddenException => err
+              raise err if retries.positive?
+              log('warn', "ManagedDisk.read: #{err} - getting new SAS URL")
+              begin
+                close
+              rescue => err
+                log('debug', "ManagedDisk.read: #{err} received on close ignored.")
+              end
+              @sas_url = @storage_service.access_token(@disk_name, @resource_group, options)
+              retries += 1
+              retry
+            end
           end
 
           def close
--- a/lib/azure/armrest/storage/managed_disk.rb
+++ b/lib/azure/armrest/storage/managed_disk.rb
@@ -1,0 +1,29 @@
+# Azure namespace
+module Azure
+  # Armrest namespace
+  module Armrest
+    # Storage namespace
+    module Storage
+      # Base class for managing managed disks.
+      module ManagedStorageHelper
+        class ManagedDisk
+          def initialize(storage_service, disk_name, resource_group, sas_url)
+            @storage_service = storage_service
+            @disk_name       = disk_name
+            @resource_group  = resource_group
+            @sas_url         = sas_url
+          end
+
+          def read(options = {})
+            @storage_service.read(@sas_url, options)
+          end
+
+          def close
+            @storage_service.close(@disk_name, @resource_group)
+            @sas_url = nil
+          end
+        end # ManagedDisk
+      end # ManagedStorageHelper
+    end # Storage
+  end # Armrest
+end # Azure
--- a/lib/azure/armrest/storage/managed_disk.rb
+++ b/lib/azure/armrest/storage/managed_disk.rb
@@ -7,11 +7,11 @@ module Azure
       # Base class for managing managed disks.
       module ManagedStorageHelper
         class ManagedDisk
-          def initialize(storage_service, disk_name, resource_group, sas_url)
+          def initialize(storage_service, disk_name, resource_group, options)
             @storage_service = storage_service
             @disk_name       = disk_name
             @resource_group  = resource_group
-            @sas_url         = sas_url
+            @sas_url         = storage_service.access_token(disk_name, resource_group, options)
           end
 
           def read(options = {})
--- a/lib/azure/armrest/storage/managed_storage_helper.rb
+++ b/lib/azure/armrest/storage/managed_storage_helper.rb
@@ -1,4 +1,5 @@
 module Azure::Armrest::Storage::ManagedStorageHelper
+  require_relative 'managed_disk'
   # Get the raw blob information for a managed disk. This is similar to
   # the StorageAccount#get_blob_raw method, but applies only to a managed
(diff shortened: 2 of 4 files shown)


lib/azure/armrest/storage/managed_disk.rb
```diff
@@ -0,0 +1,29 @@
...
+    # Storage namespace
+    module Storage
+      # Base class for managing managed disks.
+      module ManagedStorageHelper
+        class ManagedDisk
+          def initialize(storage_service, disk_name, resource_group, sas_url)
+            @storage_service = storage_service
+            @disk_name       = disk_name
+            @resource_group  = resource_group
+            @sas_url         = sas_url
```
roliveri: Can't you acquire the `sas_url` here and not have to pass it in? This way you encapsulate the `ManagedDisk` semantics in this class. Then, any needed re-acquisition of the `sas_url` can happen here as well - in the case of a failed `read`.

lib/azure/armrest/storage/managed_storage_helper.rb
```diff
@@ -49,37 +50,13 @@ module Azure::Armrest::Storage::ManagedStorageHelper
...
-    # Get the SAS URL from the BeginGetAccess call
-    op_url = headers.try(:azure_asyncoperation) || headers.location
-
-    # Dig the URL + SAS token URL out of the response
-    response = rest_get(op_url)
-
-    body = Azure::Armrest::ResponseBody.new(response.body)
-    sas_url = body.properties.output.access_sas
+  def open(disk_name, resource_group = configuration.resource_group, options = {})
+    sas_url = access_token(disk_name, resource_group, options)
```
roliveri: See my above comment on acquiring the `sas_url` in the `ManagedDisk` instantiation.

lib/azure/armrest/storage/managed_disk.rb
jerryk55: Oh, how I tried.  For two days. I could not for the life of me figure out how to get the ManagedDisk work with the modules in question and allow it to call all the methods that live up in AzureArmrestService that are required to do everything necessary.  If somebody wants to show me how to do that I am more than willing to change this all again.

lib/azure/armrest/storage/managed_storage_helper.rb
jerryk55: Ditto.

jerryk55:
@roliveri requested changes on the dependent PR to handle the ForbiddenException inside the ManagedDisk object here and that has been done.

lib/azure/armrest/storage/managed_disk.rb
roliveri: What's missing in `ManagedDisk`? You have:
```ruby
def initialize(storage_service, disk_name, resource_group, sas_url)
```
passed in to the initializer.
In the instantiating code you call:
```ruby
sas_url = access_token(disk_name, resource_group, options)
```
So you have: `disk_name` and `resource_group` needed to call `access_token()` from `storage_service`.
It seems like all you need to do is pass `options` into initialize as well.

lib/azure/armrest/storage/managed_disk.rb
```diff
@@ -0,0 +1,43 @@
+# Azure namespace
+module Azure
+  # Armrest namespace
+  module Armrest
+    # Storage namespace
+    module Storage
+      # Base class for managing managed disks.
+      module ManagedStorageHelper
+        class ManagedDisk
+          def initialize(storage_service, disk_name, resource_group, sas_url)
```
roliveri: So, this would be:
```ruby
def initialize(storage_service, disk_name, resource_group, options)
```

lib/azure/armrest/storage/managed_disk.rb
```diff
@@ -0,0 +1,43 @@
...
+            @storage_service = storage_service
+            @disk_name       = disk_name
+            @resource_group  = resource_group
```
roliveri: Add:
```ruby
@open_options = options
```

lib/azure/armrest/storage/managed_disk.rb
```diff
@@ -0,0 +1,43 @@
...
+            @sas_url         = sas_url
```
roliveri: This becomes:
```ruby
@sas_url = @storage_service.access_token(@disk_name, @resource_group, @open_options)
```

lib/azure/armrest/storage/managed_storage_helper.rb
```diff
@@ -49,37 +50,13 @@ module Azure::Armrest::Storage::ManagedStorageHelper
...
+    ManagedDisk.new(self, disk_name, resource_group, sas_url)
+  end
 
-    # The same restrictions that apply to the StorageAccont method also apply here.
+  def read(sas_url, options = {})
```
roliveri: No need to pass `options` into `read`, use `@open_options`

jerryk55: No - the options have the read offset and length.  This changes for every call to read.

jerryk55:
@roliveri that all makes sense.  Modified and tested and ready for further review.  I may need to fix tests as well still.

jerryk55:
@roliveri tests have been fixed.  This can go unless we need to wait for @djberg96 or @bronaghs to ok it.

lib/azure/armrest/storage/managed_storage_helper.rb
```diff
@@ -107,17 +83,71 @@ def get_blob_raw(disk_name, resource_group = configuration.resource_group, optio
...
-      rest_post(end_url)
+  end
+
+  def close(disk_name, resource_group)
+    end_url = build_url(resource_group, disk_name, 'EndGetAccess')
+    rest_post(end_url)
+  end
+
+  def get_blob_raw(disk_name, resource_group = configuration.resource_group, options = {})
+    sas_url = open(disk_name, resource_group, options)
```
djberg96: Confused here. This returns a `ManagedDisk.new` object, doesn't it?

jerryk55:
@djberg96 fixed up the get_blob_raw issues and tests passed if you can give another perusal...

djberg96:
👍

jerryk55:
@miq-bot add_label fine/yes

PR #358, opened 2018-02-22
Handle possibility of colons in keys for model generation.

robbmanes:
Similarly to commit 845d11a (pr 329) it is possible for a colon to be present in keys of a BaseModel.  This patch allows for colons to be converted to underscores like we do for periods and dollar signs.

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1548153

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -225,9 +225,7 @@ module Azure
       def __setobj__
         excl_list = self.class.send(:excl_list)
         @hashobj.each do |key, value|
-          snake = key.to_s.tr(' ', '_').underscore
-          snake.tr!('.', '_')
-          snake.tr!('$', '_')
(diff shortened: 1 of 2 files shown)


djberg96:
I'm wondering if we should just change it to `[\W]` at this point, i.e. anything that's not a letter or number.

djberg96:
@robbmanes There are some existing specs for testing stuff. They seem to pass, but if you wouldn't mind adding one, that would be great. You can pretty much copy this:

https://github.com/ManageIQ/azure-armrest/blob/master/spec/models/base_model_spec.rb#L217-L222

robbmanes:
Sorry, don't think I was very clear; I did in the rebase for 50d3a91 - I meant if you wanted more symbols than the ones I was testing against (added pound sign and colon) like parenthesis or anything else other than the ones I already added.

djberg96:
@robbmanes oops, nah, that should be fine.

djberg96:
👍

PR #362, opened 2018-03-09
Added the get_by_url method to StorageAccountService

djberg96:
This PR adds the `StorageAccountService#get_from_url` method. This is handy when you have a storage account endpoint like `http://foo.blob.core.windows.net/vhds/some_file.vhd` and want to get information on the storage account that's backing it. I also updated the `get_from_vm` method to take advantage of this new method since it's more generic, and more efficient.

This will be a handy method to have as part of the targeted refresh update.

### DIFF
--- a/lib/azure/armrest/storage_account_service.rb
+++ b/lib/azure/armrest/storage_account_service.rb
@@ -28,6 +28,33 @@ module Azure
         super(filter).each { |model| model.configuration = configuration }
       end
 
+      # Retrieve information about the storage account using a URI. You may optionally
+      # specify a resource group name, which will slightly faster. By default it will
+      # use the resource group specified in the service configuration, if any.
+      #
+      # Example:
+      #
+      #   sas = Azure::Armrest::StorageAccountService.new(<config>)
+      #   url = https://foo.blob.core.windows.net/vhds/dberger1201691213010.vhd"
+      #
+      #   sas.get_from_url(url)
+      #   sas.get_from_url(url, some_resource_group)
+      #
+      def get_from_url(url, resource_group = configuration.resource_group)
+        uri  = Addressable::URI.parse(url)
+        name = uri.host.split('.').first
+
+        unless resource_group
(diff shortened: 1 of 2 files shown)


lib/azure/armrest/storage_account_service.rb
```diff
@@ -28,6 +28,33 @@ def list_all(filter = {})
...
+
+        unless resource_group
+          rservice = Azure::Armrest::ResourceService.new(configuration)
+          filter   = "resourceType eq 'Microsoft.Storage/storageAccounts' and name eq '#{name}'"
+          resource = rservice.list_all(:filter => filter, :all => true).first
+          raise ArgumentError, "unable to find resource group for #{url}" unless resource
+          resource_group = resource.id[/resourceGroups\/(.*?)\//i, 1]
+        end
+
+        get(name, resource_group)
```
bronaghs: @djberg96 - Do you need to set the configuration like you did before in the `get_from_vm` method? Like this:
`get(name, resource_group).configuration = configuration` ?

djberg96: @bronaghs No, it will already have that information.

bronaghs:
@djberg96 I want to make sure I understand the logic correctly -

When the resource group is _not_ known:
Before - you would `get` the storage account directly from the StorageAccountService by looking up the account name.
After  - you use the ResourceService class to search through all the StorageAccount instances and filter on the account name? Then the resource group is parsed from the first result returned by the search.

Can you confirm I understood this part of the PR? Thanks.

djberg96:
@bronaghs Pretty close. Before we would do a `list_all` and filter by name after the fact which was slower and consumed more memory. The `ResourceService` is the one service that actually let's us filter up front, effectively reducing the result set to a single record. It returns a smaller payload to boot. So, we just use that to get the id string, and then parse the resource group out of that.

bronaghs:
Thanks @djberg96 - sounds good.

Issue #306, opened 2017-09-13
VirtualMachineImageService#list_all is very slow


djberg96:
As the title says, `VirtualMachineImageService#list_all` is very slow. It's a quadruple nested loop, starting with publishers, then offers, then skus, then versions. The problem is that the number of publishers is now over 500, so it's taking hundreds of requests and lots of time to complete.

In the current version we need to parallel these loops, or at least the outer loop, to improve performance. If we switch to the excon library, we should pipeline these requests using an approach like this:

https://gist.github.com/djberg96/f44d2cb7b930c2898ce7a8466bd54956

djberg96:
Fixed by https://github.com/ManageIQ/azure-armrest/pull/363

PR #363, opened 2018-03-12
Use parallel for nested loops in VirtualMachineImageService#list_all method

djberg96:
This updates the `VirtualMachineImageService#list_all` method to use multiple loops via the Parallel library in order to significantly speed up the results. Currently this method takes over 20 minutes without threading. With threading the time drops down to about 1 minute to collect all results.

Addresses https://github.com/ManageIQ/azure-armrest/issues/306

### DIFF
--- a/lib/azure/armrest/virtual_machine_image_service.rb
+++ b/lib/azure/armrest/virtual_machine_image_service.rb
@@ -31,23 +31,28 @@ module Azure
       def list_all(location = @location)
         raise ArgumentError, "No location specified" unless location
 
+        mutex  = Mutex.new
         images = []
-        publishers(location).each do |publisher|
-          offers(location, publisher.name).each do |offer|
-            skus(offer.name, location, publisher.name).each do |sku|
-              versions(sku.name, offer.name, location, publisher.name).each do |version|
-                images << Azure::Armrest::VirtualMachineImage.new(
+
+        Parallel.each(publishers(location), :in_threads => configuration.max_threads) do |publisher|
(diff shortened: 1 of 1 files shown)


lib/azure/armrest/virtual_machine_image_service.rb
```diff
@@ -31,23 +31,28 @@ def initialize(configuration, options = {})
...
-            skus(offer.name, location, publisher.name).each do |sku|
-              versions(sku.name, offer.name, location, publisher.name).each do |version|
-                images << Azure::Armrest::VirtualMachineImage.new(
+
+        Parallel.each(publishers(location), :in_threads => configuration.max_threads) do |publisher|
+          Parallel.each(offers(location, publisher.name), :in_threads => configuration.max_threads) do |offer|
+            Parallel.each(skus(offer.name, location, publisher.name), :in_threads => configuration.max_threads) do |sku|
+              Parallel.each(versions(sku.name, offer.name, location, publisher.name), :in_threads => configuration.max_threads) do |version|
+                mutex.synchronize do
+                  images << Azure::Armrest::VirtualMachineImage.new(
```
bzwei: Where is the place that the code needs to make API calls to the provider? Is it in the constructor? Maybe we can use multiple loops to collect the hashes, and finally use one `Parallel.each` to construct images.

djberg96: The `versions`, `skus` and `offers` and `publishers` methods are where the actual API calls happen.

PR #368, opened 2018-04-17
Add a customizable method that can disable key transformation

djberg96:
In version 0.8.2 we added the ability to allow either `lower_case` or `camelCase` options for the `create` method. This convenience was handy and consistent with our model generation methods.

However, this caused a bug in the `TemplateDeploymentService#create` method where you don't actually want to transform keys in templates, but keep them as-is, because they aren't interchangeable.

This PR fixes the issue by updating the `ResourceGroupBasedService` and `TemplateDeploymentService`. From now on it will use the private method `transform_keys?` internally to check to see if the keys should be transformed first. Individual subclasses can override this method simply by redefining it to return `false`.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1568009

Update: as per Bill's request, the method was altered so that subclasses have more granular control over which keys are transformed.

### DIFF
--- a/lib/azure/armrest/resource_group_based_service.rb
+++ b/lib/azure/armrest/resource_group_based_service.rb
@@ -48,7 +48,8 @@ module Azure
         url = build_url(rgroup, name)
         url = yield(url) || url if block_given?
 
-        body = options.deep_transform_keys{ |k| k.to_s.camelize(:lower) }.to_json
+        body = transform_create_options(options).to_json
+
         response = rest_put(url, body)
 
         headers = Azure::Armrest::ResponseHeaders.new(response.headers)
@@ -189,6 +190,13 @@ module Azure
       end
 
       private
+
+      # By default, camelize all hash options for the create method.
+      # Subclasses should override this behavior as needed.
+      #
+      def transform_create_options(hash)
+        hash.deep_transform_keys{ |k| k.to_s.camelize(:lower) }
+      end
 
       def convert_id_string_to_url(id_string, info = nil)
         if id_string.include?('api-version')
--- a/lib/azure/armrest/template_deployment_service.rb
+++ b/lib/azure/armrest/template_deployment_service.rb
@@ -74,6 +74,19 @@ module Azure
 
       private
 
+      # Don't transform templates or parameters for deployments.
+      #
+      def transform_create_options(hash)
+        properties = hash['properties'] || hash[:properties]
+        return super(hash) unless properties
+
+        ignored = properties.extract!(:template, 'template', :parameters, 'parameters')
+        hash = super(hash)
+        (hash['properties'] || hash[:properties]).merge!(ignored)
+
+        hash
+      end
+
       def delete_resources(ids, retry_cnt)
         if retry_cnt == 0
           ids.each { |id| log("error", "Failed to delete #{id}") }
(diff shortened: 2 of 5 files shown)


lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -48,7 +48,12 @@ def create(name, rgroup = configuration.resource_group, options = {})
         url = build_url(rgroup, name)
         url = yield(url) || url if block_given?
 
-        body = options.deep_transform_keys{ |k| k.to_s.camelize(:lower) }.to_json
+        if transform_keys?
```
bzwei: better `body = transform_create_options(options).to_json` here, and 

```
  def transform_create_options(options)
    options.deep_transform_keys { |k| k.to_s.camelize(:lower) }
  end
```

lib/azure/armrest/resource_group_based_service.rb
```diff
@@ -190,6 +195,13 @@ def delete(name, rgroup = configuration.resource_group)
 
       private
 
+      # Determines whether or not the calling method will perform a deep
+      # transform plus camelize on any hash options.
+      #
+      def transform_keys?
```
bzwei: For template deployment the `options` has structure [here](https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/cloud_manager/orchestration_stack.rb#L21)

```
  { 'properties' => {
    :template => ...
    :parameters => ...
    :other...
}
```
Therefore when override `transform_create_options` we can skip `properties.parameters` and continue to transform others keys.

djberg96: @bzwei shouldn't we skip the template content as well?

bzwei: The template should be a string. But yes, skip it for safety.

JPrause:
@miq-bot add_label blocker

JPrause:
@djberg96 can you add the gaprindashvili/yes label if this can be backported.

djberg96:
@JPrause This wouldn't be directly backported. I would release a new version of the gem, then submit a gemspec update to the azure provider repo. That PR would then be tagged.

lib/azure/armrest/template_deployment_service.rb
```diff
@@ -74,6 +74,16 @@ def exists?(deploy_name, resource_group = configuration.resource_group)
 
       private
 
+      # Don't transform templates or parameters for deployments.
+      #
+      def transform_create_options(hash)
+        properties = hash['properties'] || hash[:properties]
+        ignored = properties.extract!(:template, 'template', :parameters, 'parameters') if properties
+        hash = hash.deep_transform_keys{ |k| k.to_s.camelize(:lower) }
+        (hash['properties'] || hash[:properties]).merge!(ignored) if properties
```
bzwei: may it be simplified as
```
  properties = ...
  return super(hash) unless properties
  ignored = ...
  super(hash).merge!(ignored)
```

djberg96: Eh, it's not that much simpler. Plus, the ignored hash has to be merged at the right level so your code won't pass the specs.

djberg96: @bzwei ok, I did some minor cleanup, but I think that's the best I can do for now. We can always refactor later if you think there's a better way.

Issue #370, opened 2018-05-01
RFE: AzureStack -- auto-discovery of Azure endpoint URLs


bsquizz:
Instead of hard-coding URLs in the Environment, we should be able to auto-discover them given only the resource manager URL using a GET to `<ARM_URL>/metadata/endpoints?api-version=1.0`

This gives us details on the gallery, graph, and portal URLs, as well as the `active_directory_authority` and `active_directory_resource_id`

This will be helpful for AzureStack deployments where the underlying URLs can't be hard-coded. When adding the provider, all the admin would need to provide is the resource manager URL and we could take care of the rest.

See an example in Ruby here (search for `def get_active_directory_settings(armEndpoint)`):
https://docs.microsoft.com/en-us/azure/azure-stack/user/azure-stack-version-profiles-ruby

JSON returned by an azure stack deployment:
```json
{
    "galleryEndpoint": "https://portal.westus.stackpoc.com:30015/",
    "graphEndpoint": "https://graph.windows.net/",
    "portalEndpoint": "https://portal.westus.stackpoc.com/",
    "authentication": {
        "loginEndpoint": "https://login.windows.net/",
        "audiences": ["https://management.poc.avahc.com/01a7976d-5a10-475d-b39d-01993f9d7d90"]
    }
}
```

This same API is working on the public cloud too. Example JSON:
```json
{
    "galleryEndpoint": "https://gallery.azure.com/",
    "graphEndpoint": "https://graph.windows.net/",
    "portalEndpoint": "https://portal.azure.com/",
    "authentication": {
        "loginEndpoint": "https://login.windows.net/",
        "audiences": ["https://management.core.windows.net/", "https://management.azure.com/"]
    }
}
```

bsquizz:
We can discover supported API versions/API profiles using ARM as well, using `/providers/<provider type>`, for example, here's a snip of the JSON for `/providers/Microsoft.Network`:

```json
{
    "namespace": "Microsoft.Network",
    "resourceTypes": [
        {
            "resourceType": "connections",
            "locations": [
                "westus"
            ],
            "apiVersions": [
                "2015-06-15",
                "2015-05-01-preview"
            ],
            "apiProfiles": [
                {
                    "profileVersion": "2017-03-09-profile",
                    "apiVersion": "2015-06-15"
                }
            ]
        },
        {
            "resourceType": "loadBalancers",
            "locations": [
                "westus"
            ],
            "apiVersions": [
                "2015-06-15",
[...]

PR #371, opened 2018-05-02
Add an Environment.discover method

djberg96:
This adds an `Environment.discover` method that, when it receives an endpoint, will automatically retrieve the metadata that Azure provides in order to create a custom `Environment` object with its own set of custom endpoints. This environment can then be passed to the `Configuration.new` method.

Example:
```
env = Azure::Armrest::Environment.discover(:url => "https://some_endpoint")
Azure::Armrest::Configuration.new(:environment => env, ...)
```

The main purpose here is to alleviate the pain for Azure stack environments where users would have to manually create their own environments, and hard coding them might break. This could be also used for any other public environments Microsoft introduces in the future that aren't already pregenerated as well.

Addresses https://github.com/ManageIQ/azure-armrest/issues/370

### DIFF
--- a/lib/azure/armrest/environment.rb
+++ b/lib/azure/armrest/environment.rb
@@ -2,22 +2,22 @@ module Azure
   module Armrest
     class Environment
       # A list of valid keys that can be passed to the constructor
-      VALID_KEYS = [ 
-        :name,
-        :active_directory_authority,
-        :active_directory_resource_id,
-        :gallery_url,
-        :graph_url,
-        :graph_api_version,
-        :key_vault_dns_suffix,
-        :key_vault_service_resource_id,
-        :publish_settings_file_url,
-        :resource_manager_url,
-        :service_management_url,
-        :sql_database_dns_suffix,
-        :storage_suffix,
-        :traffic_manager_dns_suffix
-      ]
+      VALID_KEYS = %i[
+        name
+        active_directory_authority
+        active_directory_resource_id
+        gallery_url
+        graph_url
+        graph_api_version
+        key_vault_dns_suffix
+        key_vault_service_resource_id
+        publish_settings_file_url
+        resource_manager_url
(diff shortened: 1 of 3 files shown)


djberg96:
@bsquizz How does that look?

lib/azure/armrest/environment.rb
```diff
@@ -74,15 +74,49 @@ def initialize(options)
...
+      alias login_endpoint active_directory_authority
       alias resource_url resource_manager_url
+      alias gallery_endpoint gallery_url
+      alias graph_endpoint graph_url
+
+      # Automatically discover and create an Environment given a resource
+      # manager endpoint. The +options+ hash must include an endpoint :url key.
+      # It may also include a :name (recommended), and http proxy options.
+      #
+      def self.discover(options = {})
```
bzwei: Because `options[:url]` is a must, `options` cannot have default value `{}`.

lib/azure/armrest/environment.rb
```diff
@@ -74,15 +74,49 @@ def initialize(options)
...
+        url  = options.fetch(:url)
+        name = options[:name] || 'Custom'
+
+        uri = Addressable::URI.join(url, '/metadata/', 'endpoints')
+        uri.query = "api-version=1.0"
```
bzwei: why hardcoded `1.0`?

djberg96: That's what the docs say to use. As far as I could tell, metadata information cannot be gathered from the resource provider service, and that would require authentication first anyway.

lib/azure/armrest/environment.rb
djberg96: Ok, I'll update that.

djberg96:
@bzwei Ok, made the argument to the `discover` method mandatory.

PR #374, opened 2018-07-11
Fix and update the ResourceProviderService#list_api_versions method

djberg96:
Currently the `ResourceProviderService#list_api_versions` method is broken. At the moment it always grabs the first result from the response. I *think* I did this because originally there was only ever one service per resource, but that changed a while ago and I never updated it. That, or it's always been broken and we never noticed since it's not used internally.

So, this PR modifies the `list_api_versions` method to require both a namespace and a service name, as it wouldn't make sense without both arguments. If a valid service name cannot be found, an `ArgumentError` is raised.

This is needed as part of an effort to handle invalid api-version strings in the ManageIQ Azure provider code.

https://bugzilla.redhat.com/show_bug.cgi?id=1566613


bzwei:
Who currently calls `list_api_versions`? Will it break the backward compatibility by changing the argument list?

djberg96:
@bzwei Nothing calls it internally or in the Azure provider code, and it's broken anyway, so backwards compatibility isn't really affected.

PR #379, opened 2018-11-08
Make VM power operations return a ResponseHeader

djberg96:
Currently VM power operations return nil. This PR modifies them to return a `ResponseHeaders` object. This is more useful in that it let's users check for potential errors, as well as use the response in conjunction with our `wait` and `poll` methods.

### DIFF
--- a/lib/azure/armrest/virtual_machine_service.rb
+++ b/lib/azure/armrest/virtual_machine_service.rb
@@ -306,8 +306,11 @@ module Azure
         raise ArgumentError, "must specify name of the vm" unless vmname
 
         url = build_url(group, vmname, action)
-        rest_post(url)
-        nil
+        response = rest_post(url)
+
+        Azure::Armrest::ResponseHeaders.new(response.headers) do |headers|
+          headers.response_code = response.code
+        end
       end
     end
   end
--- a/lib/azure/armrest/virtual_machine_service.rb
+++ b/lib/azure/armrest/virtual_machine_service.rb
@@ -308,7 +308,7 @@ module Azure
         url = build_url(group, vmname, action)
         response = rest_post(url)
 
-        Azure::Armrest::ResponseHeaders.new(response.headers) do |headers|
+        Azure::Armrest::ResponseHeaders.new(response.headers).tap do |headers|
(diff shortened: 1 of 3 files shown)


lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -306,8 +306,11 @@ def vm_operate(action, vmname, group, options = {})
         raise ArgumentError, "must specify name of the vm" unless vmname
 
         url = build_url(group, vmname, action)
-        rest_post(url)
-        nil
+        response = rest_post(url)
+
+        Azure::Armrest::ResponseHeaders.new(response.headers) do |headers|
```
bzwei: do you miss `.tap`?

djberg96: Oops, fixed.

spec/virtual_machine_service_spec.rb
```diff
@@ -111,6 +111,29 @@
...
   end
 
+  context "operations" do
+    let(:response) { IO.read('spec/fixtures/operations_response.json') }
+
+    before do
+      allow(vms).to receive(:rest_post).and_return(response)
+      allow(response).to receive(:code).and_return(202)
+      allow(response).to receive(:body).and_return('')
+      allow(response).to receive(:headers).and_return(response)
```
bzwei: To avoid misleading, better to append `headers` such as 
```
  let(:response_headers) { IO.read('spec/fixtures/operations_response_headers.json') }
  ...
    allow(response).to receive(:headers).and_return(response_headers)
```
Also please assert `response_code` in headers.

spec/virtual_machine_service_spec.rb
```diff
@@ -111,6 +111,33 @@
     end
   end
 
+  context "operations" do
+    let(:response) { RestClient::Response.new }
```
bzwei: consider some simplification like
```
    let(:response_headers) { IO.read('spec/fixtures/operation_response.json') }
    let(:response) { double(:response, :code => 202, :headers => response_headers) }
    
    before { allow(vms).to receive(:rest_post).and_return(response) }
```

PR #381, opened 2018-11-13
Modify the VirtualMachineService#get method to include instance view information

djberg96:
As of June 2018 the Azure REST API was modified to allow inline expansion of the instance view properties in a standard GET call for a single instance:

https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/get

This would be handy for the Azure provider, since we currently end up making a separate call for each VM to get power state information.

Supporting the `$expand` url parameter was also necessary for this to work.

Oddly, the ability to use `$expand` does not yet appear to be supported for other calls.

Note that this is technically a breaking change and we should consider a major version bump if accepted.

### DIFF
--- a/lib/azure/armrest/virtual_machine_service.rb
+++ b/lib/azure/armrest/virtual_machine_service.rb
@@ -75,13 +75,8 @@ module Azure
       # information. If the +instance_view+ parameter is false, then only the
       # model view information will be retrieved.
       #
-      def get(vmname, group = configuration.resource_group, include_instance_view = true)
-        url = if include_instance_view
-          build_url(group, vmname, :expand => 'instanceView')
-        else
-          build_url(group, vmname)
-        end
-
+      def get(vmname, group = configuration.resource_group, options = {})
+        url = build_url(group, vmname, options)
         response = rest_get(url)
         VirtualMachineInstance.new(response)
       end
--- a/lib/azure/armrest/virtual_machine_service.rb
+++ b/lib/azure/armrest/virtual_machine_service.rb
@@ -86,7 +86,7 @@ module Azure
       # information.
       #
       def get_model_view(vmname, group = configuration.resource_group)
-        get(vmname, group, false)
+        get(vmname, group)
       end
 
       # Convenient wrapper around the get method that retrieves only the
--- a/lib/azure/armrest/virtual_machine_service.rb
+++ b/lib/azure/armrest/virtual_machine_service.rb
@@ -71,23 +71,31 @@ module Azure
       # Retrieves the settings of the VM named +vmname+ in resource group
       # +group+, which will default to the same as the name of the VM.
       #
-      # By default this method will retrieve the model view. If the +model_view+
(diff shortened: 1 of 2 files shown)


bzwei:
To avoid the need of a new major version, can we keep every untouched but add a new method called `get_model_and_instance_view`?

djberg96:
@bzwei I don't really want to add another method. Instead I changed it to just add an optional 3rd argument, so it won't be included by default, but you could do this:

`vm.get(vm_name, vm_group, :expand => 'instanceView')`

This gives us the capability without breaking backwards compatibility.

lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -71,23 +71,26 @@ def generalize(vmname, group = configuration.resource_group)
...
-      # By default this method will retrieve the model view. If the +model_view+
-      # parameter is false, it will retrieve an instance view. The difference is
-      # in the details of the information retrieved.
+      # You can also specify any query options. At this time only the
+      # :expand => 'instanceView' option is supported, but others could
+      # be added over time.
       #
-      def get(vmname, group = configuration.resource_group, model_view = true)
-        model_view ? super(vmname, group) : get_instance_view(vmname, group)
+      def get(vmname, group = configuration.resource_group, options = {})
```
bzwei: the argument list still changed, which breaks the compatibility.

lib/azure/armrest/virtual_machine_service.rb
```diff
@@ -71,23 +71,26 @@ def generalize(vmname, group = configuration.resource_group)
...
       end
 
       # Convenient wrapper around the get method that retrieves the model view
-      # for +vmname+ in resource_group +group+.
+      # for +vmname+ in resource_group +group+ without the instance view
+      # information.
       #
       def get_model_view(vmname, group = configuration.resource_group)
-        get(vmname, group, true)
+        get(vmname, group, false)
```
bzwei: the argument list does not match the new signature.

lib/azure/armrest/virtual_machine_service.rb
djberg96: Eh, it doesn't break existing behavior, though, since the 3rd argument defaults to empty.

bzwei: It breaks. consider a call with `get(vm, group, false)`. The rest call may fail, or least not returning what it used to.

djberg96: Ah, true. I'd rather break the api than add another method, though, because this is more flexible going forward.

bzwei: With a minor effort you can at least check whether `options` is a `Hash`. If not, treat it as a boolean.

djberg96: @bzwei Ok, updated.

bzwei (approved):

PR #382, opened 2019-01-11
Memoize the ResourceProviderService#supported? method

djberg96:
Let's memoize the `ResourceProviderService#supported?` since that value is unlikely to change.


bzwei:
Do we need to occasionally refresh the memorized values for a long run client?

djberg96:
@bzwei In theory a service could be registered after the fact, but I think it's pretty rare in practice. They can always bounce the client, yes.

PR #384, opened 2019-01-11
Fix custom inspect method spec

djberg96:
We can't rely on attributes appearing in a specific order for our custom inspect method, so this PR modifies the spec so that it just checks for their presence.


djberg96:
@bzwei Right, we need to remove 2.2.x from the matrix. Please ignore the 2.2.x failure since it's a bundler/travis issue.

PR #386, opened 2019-10-22
Allow memoist 0.16.0

jrafanie:


azure-armrest.gemspec
```diff
@@ -19,7 +19,7 @@ behind the scenes.
 
   spec.add_dependency 'json', '~> 2'
   spec.add_dependency 'rest-client', '~> 2.0.0'
-  spec.add_dependency 'memoist', '~> 0.15.0'
+  spec.add_dependency 'memoist', '>= 0.15.0'
```
Fryguy: Would `~> 0.15` be better so as not to bring in a 1.0 it that ever happens?

jrafanie: good idea

jrafanie:
@Fryguy thanks, now it's a straight deletion of 2 characters.  Good work!

PR #387, opened 2020-01-22
Support data disk deletion when deleting associated resources

djberg96:
Currently the `VirtualMachineService#delete_associated_resources` method can delete the OS disk, but it cannot delete attached disks (i.e. data disks). This PR adds a `data_disks` option, which is false by default. If true it will iterate over all attached data disks and delete them.

Needed for https://bugzilla.redhat.com/show_bug.cgi?id=1788053

For testing, I created three VM's in the Azure dev environment, one with no attached disks, one with 1 attached disk and one with 2 attached disks. All 3 were deleted successfully.


djberg96:
@miq-bot remove_label wip

Issue #391, opened 2020-02-24
Metrics collection needs to be updated


djberg96:
At some point the REST API for metrics collection changed. Now, instead of just a generic `filter` parameter, it now uses (and indeed requires) several more URI parameters that are used to filter the query results:

https://docs.microsoft.com/en-us/rest/api/monitor/metrics/list

The current code technically still works since it still accepts a `filter` option, but will generally fail if you try to use any of the other URI parameters within that filter.

djberg96:
Relevant PR was merged, closing.

PR #392, opened 2020-02-24
Update the list_metrics method to use the updated API

djberg96:
A while back the Microsoft API for metrics collection changed. It used to be that everything had to be done through a `filter` parameter. However, that has changed, and their REST API now accepts several other URI parameters.

https://docs.microsoft.com/en-us/rest/api/monitor/metrics/list#resulttype

The current code, while technically still functional, is no longer particularly useful. This is because in many cases the old filters will no longer work, and will typically be rejected outright. For example, a query that was once written like this will no longer work:

```
filter = "(name.value eq 'Percentage CPU' or name.value eq 'Disk Read Bytes') "
filter << "and startTime eq 2020-02-23 and endTime eq 2020-02-24"
metrics = mets.list_metrics(vm, filter)
```

Using the new API, it would be written like this:
```
options = {
  :timespan    => "2020-02-23T09:23:42-05:00/2020-02-24T09:23:42-05:00",
  :metricnames => 'Percentage CPU'
}

metrics = mets.list_metrics(vm, options)
```

So this PR changes the second parameter of the `list_metrics` method to a hash of options.

Currently we've had to write custom code in the ManageIQ Azure provider code to deal with this:

https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/cloud_manager/metrics_capture.rb#L273

Addresses https://github.com/ManageIQ/azure-armrest/issues/391

### DIFF
--- a/lib/azure/armrest/insights/metrics_service.rb
+++ b/lib/azure/armrest/insights/metrics_service.rb
@@ -36,9 +36,21 @@ module Azure
           )
         end
 
-        # Returns a list metrics for +resource_id+, which can be
-        # either a resource object or a plain resource string. You
-        # may also provide a +filter+ to limit the results.
+        # Returns a list metrics for +resource_id+, which can be either a
+        # resource object or a plain resource string. You may also provide
+        # hash of filtering +options+ to limit the results. The possible
+        # options are:
+        #
+        #   * :timespan        => The timespan of the query in "start/end" format.
+        #   * :interval        => The interval (timegrain) of the query.
+        #   * :metricnames     => A comma separated list of metrics to retrieve.
+        #   * :aggregation     => A comma separated list of aggregration types to retrieve.
+        #   * :segment         => The name of the dimension to segment the metric values by.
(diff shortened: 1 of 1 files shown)


lib/azure/armrest/insights/metrics_service.rb
```diff
@@ -64,7 +80,12 @@ def list_metrics(resource, filter = nil)
           )
 
           url << "?api-version=#{api_version}"
-          url << "&$filter=#{filter}" if filter
+
+          # The :filter option requires a leading '$'
+          options.each do |key, value|
+            key.to_s == 'filter' ? url << "&$" : url << "&"
```
bzwei: Does it maintain the backward compatibility? The old code directly accepts the value of the filter without key `filter`.

djberg96: @bzwei It doesn't. I guess I *could* keep backwards compatibility for a time and check to see if `options` is a string instead of a hash.

djberg96: @bzwei Ok, updated.

PR #396, opened 2020-05-22
Use manageiq style

bdunne:
Depends on: https://github.com/ManageIQ/miq_bot/pull/506


djberg96:
@bdunne Should we update activesupport while we're here?

bdunne:
> @bdunne Should we update activesupport while we're here? I'm thinking we should match core with: `"~> 5.2.4", ">= 5.2.4.3"`.

manageiq-style doesn't have a dependency on activesupport.  I think that should be a separate PR.

bdunne:
@miq-bot assign Fryguy

PR #397, opened 2020-10-19
Update default ssl_version to TLSv1_2

djberg96:
Set the default `ssl_version` config option to TLS 1.2 as per the recommendations from Microsoft: https://azure.microsoft.com/en-us/updates/azuretls12/


djberg96:
@bzwei Please review. Note that the Hakiri failure is unrelated.

PR #399, opened 2020-11-19
[WIP] Add a VirtualMachineService#skus method

djberg96:
The original method to get flavor information using the `VirtualMachineService#series` method has been deprecated by microsoft in favor of a "skus" call. This was partly inspired by a user that ran into a provisioning error when a particular flavor was selected, but wasn't actually supported for the region because of generation v1 vs v2 support, resulting in an error.

Although technically other resources can be pulled via this call, I put it under VirtualMachineService since it's (somewhat strangely) in the Microsoft.Compute namespace instead of Microsoft.Resources.

Ultimately the idea here is to include more information for the flavors in ManageIQ.

See also:

https://docs.microsoft.com/en-us/rest/api/compute/resourceskus/list

https://docs.microsoft.com/en-us/rest/api/compute/virtualmachinesizes/list

EDIT: Actually, I think this should be its own service. Marking as WIP for now.

### DIFF
--- a/lib/azure/armrest/model/base_model.rb
+++ b/lib/azure/armrest/model/base_model.rb
@@ -305,6 +305,7 @@ module Azure
     class VirtualMachineExtension < BaseModel; end
     class VirtualMachineImage < BaseModel; end
     class VirtualMachineSize < BaseModel; end
+    class VirtualMachineSku < BaseModel; end
 
     module HDInsight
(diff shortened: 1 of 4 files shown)


mikebutak:
Thanks for this quick fix, Daniel!

djberg96:
Gonna have to shelve this for now. I definitely think it should be its own service.

PR #400, opened 2020-11-20
Fix tenants method

djberg96:
There is no `configuration.resource_url` method. It's supposed to be `configuration.environment.resource_url`.


djberg96:
@miq-bot add_reviewer @bzwei

bzwei (approved):

PR #401, opened 2020-12-10
Internal StorageAccount model fixes, other fixes

djberg96:
Running the rubocop linter over the code base revealed a bug in the `StorageAccount` model where the `transform_keys` method wasn't actually doing anything. I don't think this showed up because typically the defaults are used.

Additionally, a duplicate exception was corrected, and `options` are now actually passed to the `vm_operate` method, which could potentially affect the `capture` method.

Lastly, a redundant require statement and some useless assignments were removed.

### DIFF
--- a/azure-armrest.gemspec
+++ b/azure-armrest.gemspec
@@ -17,7 +17,7 @@ different than the current azure gem, which uses the older (XML) interface
 behind the scenes.
   EOF
 
-  spec.add_dependency 'activesupport', '>= 4.2.2'
+  spec.add_dependency 'activesupport', '~> 5.2'
   spec.add_dependency 'addressable', '~> 2.5.0'
   spec.add_dependency 'azure-signature', '~> 0.3.0'
   spec.add_dependency 'json', '~> 2'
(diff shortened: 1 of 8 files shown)


djberg96:
@bzwei Please review.

djberg96:
Seems I had to update the activesupport dependency, otherwise `1.month` breaks with newer versions.

NickLaMuro:
@djberg96 We are going to need a Rails 6.0 compatible version of this then because I have been working on that for the past couple of months, and releasing something like this will probably break the build:

https://github.com/ManageIQ/manageiq/issues/19977

We are getting very close to releasing that as well.

djberg96:
@NickLaMuro Ok, updated. Near as I can tell that works with Rails 5 or 6.

NickLaMuro:
@djberg96 awesome, thanks!  I was going to look into doing tackling either something like you did here or an alternative `rails-6` branch for this same work, but you beat me to it!

Thanks again!

NickLaMuro:
That said, it might not hurt to rebase down that last to commits, and then extract that commit to a new branch and do a release of the gem that fixes the support (since this is will be failing on master).

Actually, I will save you the trouble and throw that PR together for you.

bzwei (approved):

bzwei:
@djberg96 can it be merged, or you'll need to do what @NickLaMuro suggested?

djberg96:
@bzwei It can be merged.

PR #402, opened 2020-12-11
Fix activesupport to for 6.1.0

NickLaMuro:
Extraction of a few commits from https://github.com/ManageIQ/azure-armrest/pull/401

Add explicit require so that `1.month` works with Rails 6.1.x

Probably could release this on it's own as a minor release.


NickLaMuro:
Feel free to close if you would rather do all of this in your previous PR.  I admittedly kinda misread the intent of #401 and thought it was doing more than `rubocop` stuff.

That said, this might be slightly less risky for a patch update, but I could see it going either way.

Anyway, you are free to do with this as you wish.

djberg96:
@NickLaMuro I appreciate it, but I think I'd rather just do it all at once in #401. It's just a bunch of miscellaneous fixes, I see no reason to split the PR.

PR #403, opened 2021-01-04
remove test reporter dependency

d-m-u:
~~otherwise i think local rubocop will break:~~

```
Bundler could not find compatible versions for gem "simplecov":
  In Gemfile:
    simplecov

    codeclimate-test-reporter (~> 1.0.0) was resolved to 1.0.9, which depends on
      simplecov (<= 0.13)
```

which is moot if we remove the dependency anyway, thanks @djberg96

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,9 @@ matrix:
   - rvm: ruby-head
   - rvm: jruby-head
   fast_finish: true
-after_script: bundle exec codeclimate-test-reporter
+before_script:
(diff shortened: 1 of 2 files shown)


djberg96:
I think codeclimate-test-reporter is deprecated and we've been removing it from other repos. Lemme check.

Fryguy:
Technically we need to add more to the .travis.yml to enable it the "new" way, but I have to find the snippet.  Gotta get caught up on stuff first :)

d-m-u:
marking as work in progress cause it needs whatever jason just said

Fryguy:
I think what you have is correct...now I just need to add a token on the Travis side, which, unfortunately, only I can do.

Issue #405, opened 2021-04-20
Azure-armrest fails to properly handle application/xml exceptions


agrare:
Some Azure endpoints (notably the storage_account model) return a response with Content-Type `application/xml` which is then parsed with `Hash.from_xml(response.body)`.

If the API call returns an error code the exception builder tries to parse the payload as JSON and if that fails takes the string directly which isn't always a valid utf-8 string.

```
RestClient.post "https://management.azure.com/subscriptions, "", "Accept"=>"application/json", "Authorization"=>"Bearer [FILTERED] ", "Content-Length"=>"0", "Content-Type"=>"application/json", "User-Agent"=>"rest-client/2.1.0 (linux-gnu x86_64) ruby/2.6.3p62"
# => 200 OK | application/json 288 bytes, 0.11s
RestClient.get "https://csb10037ffea12a499b.blob.core.windows.net/?comp=list", "Accept"=>"*/*", "User-Agent"=>"rest-client/2.1.0 (linux-gnu x86_64) ruby/2.6.3p62", "auth_string"=>"true", "authorization"=>"SharedKey [FILTERED] ", "content-type"=>"", "x-ms-date"=>"Sat, 03 Apr 2021 17:28:04 GMT", "x-ms-version"=>"2016-05-31"
# => 400 BadRequest | application/xml 266 bytes, 0.09s
```

Which ends up returning an XML document:
```
<?xml version="1.0" encoding="utf-8"?>
<Error>
  <Code>TlsVersionNotPermitted</Code>
  <Message>The TLS version of the connection is not permitted on this storage account.
RequestId:a1b7c24b-101e-0136-2fb2-28cc51000000
Time:2021-04-03T17:55:47.6769041Z
  </Message>
</Error>
```

But with a leading Byte-Order-Mark: `xEF\xBB\xBF<?xml version=\"1.0\"`

PR #406, opened 2021-04-20
Handle application/xml exception content-types

agrare:
If an Azure REST API call returns an error response with a content-type of application/xml then trying to JSON.parse() or err.try(:message) is leading to invalid exception message strings being returned.

Fixes https://github.com/ManageIQ/azure-armrest/issues/405

### DIFF
--- a/lib/azure/armrest/armrest_service.rb
+++ b/lib/azure/armrest/armrest_service.rb
@@ -1,5 +1,6 @@
 require 'time'
 require 'active_support/core_ext/hash/conversions'
+require 'active_support/core_ext/hash/keys'
 require_relative 'model/base_model'
 
 module Azure
@@ -253,11 +254,22 @@ module Azure
 
         def raise_api_exception(err)
           begin
-            response = JSON.parse(err.http_body)
-            code     = response['error']['code']
-            message  = response['error']['message']
+            content_type_header = err.response.headers[:content_type]
+
+            response = case content_type_header.match(%r{(application/\w+)})[1]
+                       when "application/json"
+                         JSON.parse(err.http_body)
+                       when "application/xml"
+                         # The XML document that is returned has Error, Code, and Message
+                         # so we need to downcase the keys to stay consistent with the
+                         # parsed-json hash.
+                         Hash.from_xml(err.http_body).deep_transform_keys(&:downcase)
+                       end
+
+            code    = response['error']['code']
+            message = response['error']['message']
           rescue
-            code = err.try(:http_code) || err.try(:code)
+            code    = err.try(:http_code) || err.try(:code)
             message = err.try(:http_body) || err.try(:message)
           end
 


lib/azure/armrest/armrest_service.rb
```diff
@@ -252,13 +252,20 @@ def rest_head(options)
...
+            message  = response['Error']['Message']
+          else
+            begin
+              response = JSON.parse(err.http_body)
+              code     = response['error']['code']
+              message  = response['error']['message']
+            rescue
+              code = err.try(:http_code) || err.try(:code)
+              message = err.try(:http_body) || err.try(:message)
+            end
```
agrare, line 268: So I'm trying to minimize the behavioral change here but ideally I think this would look like:
```
case
err.response.headers[:content_type]
when "application/json"
  response = JSON.parse(err.http_body)
  code     = response['error']['code']
  message  = response['error']['message']
when "application/xml"
  response = Hash.from_xml(err.http_body)
  code     = response['Error']['Code']
  message  = response['Error']['Message']
else
  code = err.try(:http_code) || err.try(:code)
  message = err.try(:http_body) || err.try(:message)
end
```

This would require some changes to the specs though as there isn't normally a non-nil `err.response` in the RestClient exception that is built

agrare: Okay I combined a bit of the switch on the content type with the rescue to catch any errors, seems like a safe-enough approach so I'm going to resolve this

djberg96:
Looks good to me. I'm amazed it took 6 years to hit this.

lib/azure/armrest/armrest_service.rb
```diff
@@ -253,11 +254,20 @@ def rest_head(options)
 
         def raise_api_exception(err)
           begin
-            response = JSON.parse(err.http_body)
-            code     = response['error']['code']
-            message  = response['error']['message']
+            response = case err.response.headers[:content_type]
+            when "application/json"
```
agrare, line 258: Some response headers have Content-Type `"application/json; charset=utf-8"` so we'll need to handle that here

agrare:
@miq-bot cross-repo-tests manageiq-providers-azure

lib/azure/armrest/armrest_service.rb
```diff
@@ -253,11 +254,22 @@ def rest_head(options)
 
         def raise_api_exception(err)
           begin
-            response = JSON.parse(err.http_body)
-            code     = response['error']['code']
-            message  = response['error']['message']
+            content_type_header = err.response.headers[:content_type]
+
+            response = case content_type_header.match(%r{(?<application_type>application\/\w+)})[:application_type]
```
Fryguy, line 259: Oddly, I think this reads better:
```suggestion
            response = case content_type_header.match(%r{(application\/\w+)})[1]
```

Fryguy: Also, if you're using %r, then there's no reason to escape the `/`

```suggestion
            response = case content_type_header.match(%r{(application/\w+)})[1]
```

OR

```suggestion
            response = case content_type_header.match(/(application\/\w+)/)[1]
```

lib/azure/armrest/armrest_service.rb
```diff
@@ -253,11 +254,22 @@ def rest_head(options)
...
+                       when "application/json"
+                         JSON.parse(err.http_body)
+                       when "application/xml"
+                         # The XML document that is returned has Error, Code, and Message
+                         # so we need to downcase the keys to stay consistent with the
+                         # parsed-json hash.
+                         Hash.from_xml(err.http_body).deep_transform_keys(&:downcase)
```
Fryguy, line 266: Hash.from_xml I believe comes from active support, so we may want to explicitly required that core extension to ensure it's loaded when we get here.

agrare: `active_support/core_ext/hash/conversions` is already required which is what providers `from_xml`: https://github.com/ManageIQ/azure-armrest/pull/406/files#diff-4c3805a9a8d1d2c792e53c71498d5c8c0354e496069340251635f99ea606d90aR2
I added `active_support/core_ext/hash/keys` for `deep_transform_keys`

Fryguy (approved):

agrare:
Released v0.11.0 with this change

PR #409, opened 2021-08-11
Add Microsoft.DBforMySQL services

agrare:
Add support for the `Azure Database for MySQL server` servers and databases endpoints


agrare:
https://rubygems.org/gems/azure-armrest/versions/0.13.0 released with this change

PR #410, opened 2021-08-11
Add Microsoft.DBforMariaDB services

agrare:
Adds support for the `Azure Database for MariaDB server` server and database endpoints


agrare:
NOTE I'll have to rebase whichever one of these gets merged second but wanted to split up each service into their own PR

